PR #57, opened 2022-01-25
docs(dev): add tokens WIP intro page

### LINKED ISSUE #56: [Docs?]: add tokens package info
### Latest version

### Summary 💡

Update `tokens` package api info. Should also put what is needed in mobile usage.

### Motivation 🔦

_No response_

caseybaggz:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

closes #56 

## What is the new behavior?
Adds intro page to tokens package docs.


website/docs/development/packages/tokens/intro.md
```diff
@@ -0,0 +1,72 @@
...
+
+The `tokens` package is the singe source of truth for all style properties and assets in the design system. This is how we define styles in `components` and why you can just use this package if you need maximum flexibility - or are on a mobile platform.
+
+Depending on what platform you are using, will determine what you are able to consume from the `tokens` package.
+
+### Web
+
+For the web (npm), we will ship `css` properties, `scss` variables, and `js` es6 modules. This way, you can choose whatever syntax works best for your project.
+
+The `components` package will consume the `css` and `js` tokens internally and will be what we return in the functions.
```
Rykus0, line 38: I think this could use a little more clarification - unless my understanding is wrong.  
Right now it sounds like the `components` package both consumes and returns the tokens.  I thought it was consuming the tokens to return attributes that can be applied to HTML elements.

website/docs/development/packages/tokens/intro.md
```diff
@@ -0,0 +1,72 @@
...
+
+#### CSS
+
+When you use the `css`, we use the `:root` psuedo-selector to store the properties.
```
Rykus0, line 42: The examples will show it, but it might improve clarity to note that the token values are expressed using CSS custom properties (variables).

website/docs/development/packages/tokens/intro.md
```diff
@@ -0,0 +1,72 @@
...
+
+:::caution
+
+This is **unreleased** documentation for TVA **tokens** package.
+
+:::
+
+#### Learn how to use TVA tokens for any project or OS platform.
+
+## npm
```
Rykus0, line 16: Should this be moved under "Web"? And maybe renamed npm/yarn, or maybe more generic, like 'dependency management'?

Rykus0 (commented):
I think we should include a section that explains the naming convention as well (once we've defined it)

website/docs/development/packages/tokens/intro.md
```diff
@@ -0,0 +1,72 @@
...
+npm install @pluralsight/tva-tokens
+```
+
+## What is a token?
+
+A "token" is a word used to define a style property for any platform. So, instead of having to specify CSS variable, iOS/Swift style, Android style variable, etc. We can just say "token" which refers to the style variables of any platform.
+
+## Usage
+
+The `tokens` package is the singe source of truth for all style properties and assets in the design system. This is how we define styles in `components` and why you can just use this package if you need maximum flexibility - or are on a mobile platform.
```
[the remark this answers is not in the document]
caseybaggz: done.

website/docs/development/packages/tokens/intro.md
caseybaggz: depends on the function you call (css vs. JS one). Let me add a little more in here that explains that further. 👍

website/docs/development/packages/tokens/intro.md
caseybaggz: That's what the code block below this shows. 😄  We create the properties and the user consumes them (using the syntax you are probably thinking of with the `var` function).

website/docs/development/packages/tokens/intro.md
caseybaggz: This is the common pattern used throughout our docs and used in most all popular docs. Installation commands are first and always use the keyword `npm` to refer to npm or Yarn. 

People are smart enough to know it refers to both especially when we have a tab to choose right below it. 😄

Rykus0 (approved):

Issue #108, opened 2022-03-08
[Feature?]: add getButtonProps & getJSButtonProps to headless-styles


caseybaggz:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

* Add CSS based `getButtonProps` and CSS-in-JS `getJSButtonProps`
* Add Codesandbox examples for vanilla, TS, styled-components, and MUI
* Add live examples into Basic Button section with nice tab component

PR #116, opened 2022-03-15
feat: add `getJSButtonProps` to Button API

caseybaggz:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

contributes to #108 

## What is the new behavior?
* Adds initial `getJSButtonProps` API
* Updates docs to match API

## Other information
There is no processor yet for sharing styles from CSS helpers to JS as I want to make sure how everything scales and works first.


packages/headless-styles/src/components/Button/buttonJS.ts
```diff
@@ -0,0 +1,272 @@
...
+}
+
+function getPsuedoStyles(kind: Kind) {
+  switch (kind) {
+    case 'text-weak':
+    case 'medium':
+    case 'strong':
+      return {
+        css: `
+          &:hover {
```
Rykus0, line 182: Can we make the assumption that they will parse these values with a parser that understands this syntax, since it's not standard CSS?  Or maybe it's just enough to indicate the syntax so they can plan appropriately?

caseybaggz: @Rykus0 yes, this is to provide for the 2 standard approaches to CSS-in-JS: stringified CSS (i.e. styled components, etc.) and CSS objects (every other tool in the world). I think once you start getting more experience of the frameworks you'll start piecing these puzzles together. 😄 

Also, when I make the examples, it should help too (next step).

packages/headless-styles/src/components/Button/buttonJS.ts
```diff
@@ -0,0 +1,272 @@
...
+
+export function getJSButtonProps(options?: ButtonOptions) {
+  const { kind, size } = getDefaultOptions(options)
+  const kindStyles = getKindStyles(kind)
+  const sizeStyles = getSizeStyles(size)
+  const psuedoStyles = getPsuedoStyles(kind)
+
+  return {
+    cssProps: `
+      ${baseCSSProps}
```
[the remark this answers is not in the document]
caseybaggz: I'm not understanding which pattern you are referring to? "base" is using the same pattern as the CSS but this is a different format?

packages/headless-styles/tests/button/buttonJS.test.ts
```diff
@@ -0,0 +1,80 @@
...
+        styles: {
+          ...xsButtonStyles,
+        },
+      })
+      expect(getJSButtonProps({ size: 's' })).toMatchObject({
+        ...defaultValue,
+        cssProps: sButtonCSS,
+        styles: { ...sButtonStyles },
+      })
+      expect(getJSButtonProps({ size: 'l' })).toMatchObject({
```
Rykus0, line 73: Should we test 'm' as well?

caseybaggz: It's already tested in the "default" one. 😉

packages/headless-styles/src/components/Button/buttonJS.ts
```diff
@@ -0,0 +1,272 @@
...
+      ${kindStyles.css.trim()}
+      ${sizeStyles.css.trim()}
+      ${psuedoStyles.css.trim()}
+    `
+      .trim()
+      .replace(/^ {2,12}/gm, ''),
```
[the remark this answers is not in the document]
Rykus0: 😄 I just wasn't sure if there was a reason to only go up to 12.  If it's condensing all leading spaces to a single, we could make a small change to cover cases where it goes above 12, but otherwise no reason to change it (other than possibly preventing this questions again)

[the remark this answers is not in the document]
Rykus0: 😄 I just wasn't sure if there was a reason to only go up to 12.  If it's condensing all leading spaces to a single, we could make a small change to cover cases where it goes above 12, but otherwise no reason to change it (other than possibly preventing this questions again)

packages/headless-styles/src/components/Button/buttonJS.ts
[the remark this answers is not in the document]
Rykus0: ah, no problem.  I was referring to the other style sets on the following lines that have js/css properties.  For the base styles you have `baseCssProps` for CSS and `buttonStyles` for JS.  
```
      ${kindStyles.css.trim()}
      ${sizeStyles.css.trim()}
      ${psuedoStyles.css.trim()}
```

packages/headless-styles/tests/button/buttonJS.test.ts
Rykus0: Ok.  Wasn't sure if there was value in testing when it's set explicitly.

packages/headless-styles/src/components/Button/buttonJS.ts
Rykus0: My concern is that we're labeling it CSS, but it's not *technically* CSS syntax.  So I want to avoid causing issues based on an expectation of standard CSS given that naming.  If it is a safe assumption to make, given the context, then I guess it's okay.

packages/headless-styles/src/components/Button/buttonJS.ts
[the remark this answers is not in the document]
caseybaggz: Ah yes, for that it's to match each design pattern (i.e. tool/preprocessor). I see what you are saying though. Make a single var with the same pattern for base. Let's worry about that when we see if we can add any preprocessors. 👍 

I like that concept though so if anything we will for sure add it in in the future.

Rykus0 (approved):

PR #134, opened 2022-03-23
refactor(hs): update styles to use tech scaling vs all in one approach

caseybaggz:
…riding

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

contributes to #108 

## What is the new behavior?
Adding chakra styles is creating recursive bugs that break MUI styles so we are adding a `tech` option to the params in Button to help control the flow of styling especially since each framework uses different approaches, this should help long term.


packages/headless-styles/src/components/Button/buttonCSS.module.css
```diff
@@ -1,20 +1,24 @@
...
     sans-serif;
-  font-size: 16px !important;
-  font-weight: inherit !important;
-  height: initial !important; /* this is due to Chakra */
-  outline-offset: initial !important;
+  font-size: 16px;
+  font-weight: 400;
+  justify-content: center;
+  outline: none;
   padding: 10px 16px;
```
Rykus0, line 15: If we use `em` for font size, it's usually good to use it for the padding so it scales proportionately to the font size

packages/headless-styles/src/components/Button/buttonCSS.ts
```diff
@@ -3,11 +3,12 @@ import type { ButtonOptions, ButtonType } from './types'
...
 
 export function getButtonProps(options?: ButtonOptions) {
-  const { kind, size } = getDefaultOptions(options)
+  const { kind, size, tech } = getDefaultOptions(options)
   const formattedSize = `size-${size}`
+  const framework = styles[tech] ?? ''
 
   return {
-    className: `ps-btn ${styles.base} ${styles[kind]} ${styles[formattedSize]}`,
+    className: `ps-btn ${styles.base} ${styles[kind]} ${styles[formattedSize]} ${framework}`,
```
Rykus0, line 11: I don't know if it helps, but we could use `Array.join(' ')` on these so we don't have to account for the possible trailing space.

Rykus0 (commented):
Just a couple questions.  Happy to see the `!important`s corralled!

packages/headless-styles/src/components/Button/buttonCSS.ts
caseybaggz: For classes, I don't think it's a big deal to have an extra space. This syntax is easier to read and faster in performance vs. forcing an Array, then adding a method on it just because there is an extra space at the end. 

Little things add up.

packages/headless-styles/src/components/Button/buttonCSS.module.css
```diff
@@ -1,20 +1,24 @@
...
+  border-radius: 6px;
   cursor: pointer;
+  display: inline-flex;
   font-family: 'PS TT Commons Roman', 'Gotham SSm A', 'Gotham SSm B', Arial,
     sans-serif;
-  font-size: 16px !important;
-  font-weight: inherit !important;
-  height: initial !important; /* this is due to Chakra */
-  outline-offset: initial !important;
+  font-size: 16px;
```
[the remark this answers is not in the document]
caseybaggz: Not opposed to this. I was honestly just using the same pattern as before, however, now that I'm testing the PS product it fails pretty miserably on the font-scaling a11y test. 😭 I would say we probably need `em` for the root `html` and then for sanity reasons, `rem` for the rest.

This unfortunately also impacts the normalize file. I'll update accordingly.

packages/headless-styles/src/components/Button/buttonCSS.module.css
caseybaggz: If we are going down this route, I think it's a little more complex than just using "r/em vomit". Each unit was created for a purpose - r/em are typography based, not spacing based. Thus, this design was a "hack" created. So, if we do change our ways, we should be more meaningful with the units.

Maybe only use rem for typography and height/width and pxs for padding/margin? I'm noting this pattern is adopted by the chakra team which so far from my testing is the best solution regarding a11y and custom font scaling.

Rykus0 (approved):

PR #138, opened 2022-03-24
docs: flesh out icon intro docs

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes


website/docs/development/packages/icons/intro.md
```diff
@@ -10,3 +10,111 @@ tags: [Development, Usage, Intro, Icons]
 This is **unreleased** documentation for Pluralsight Design **icons** package.
 
 :::
+
+The `icons` package provides icons ready to insert into your application as inline SVG elements.
```
caseybaggz, line 14: Let's revise this intro line to use the same wording and styling as the "tokens > intro.md" for consistency.

website/docs/development/packages/icons/intro.md
```diff
@@ -10,3 +10,111 @@ tags: [Development, Usage, Intro, Icons]
...
+
+## npm
+
+To use icons in your application, first install the npm package.
```
caseybaggz, line 18: Same for this line. Consistency is key. 😉

website/docs/development/packages/icons/intro.md
```diff
@@ -10,3 +10,111 @@ tags: [Development, Usage, Intro, Icons]
...
+
+```bash npm2yarn
+npm install @pluralsight/icons@alpha
+```
+
+## Usage
```
caseybaggz, line 24: I'm not seeing a reason to include a Usage heading here? Let's remove it and change all the other headings to use the h2 type (what usage does).

website/docs/development/packages/icons/intro.md
```diff
@@ -10,3 +10,111 @@ tags: [Development, Usage, Intro, Icons]
...
+
+### Inline SVG Elements
```
caseybaggz, line 26: So, what I'm seeing here overall is not the correct usage of an "intro" page. Instead, you are using it to define an API. So instead, Let's move the majority of the content of this page into another called "Icons" where that is responsible for the API.

Then, it free's up this page for you to add more content _about_ our philosophy around icons, what people can expect from the package, why, etc. Once again, take your influence from the Tokens > intro page.

From what I'm seeing here - you are just jumping right into the API assuming people will automatically know the context of why we are delivering what we are. They will not.

Assume you are writing these docs for a junior engineer who has never used an icons package or built anything with icons. We need much more information and context.

caseybaggz (commented):
Let's re-approach this by adding a more in-depth intro guide that matches the same depth tokens > intro uses and taking some of your RFC context.

website/docs/development/packages/icons/api.md
```diff
@@ -0,0 +1,223 @@
...
+
+```bash npm2yarn
+npm install @pluralsight/icons@alpha
+```
+
+## Usage
+
+:::note
+
+This library provides only the icon element to allow for maximum flexibility.
```
caseybaggz, line 26:
```suggestion
This library provides only the svg element to allow for maximum flexibility.
```

website/docs/development/packages/icons/api.md
```diff
@@ -0,0 +1,223 @@
...
+
```
caseybaggz, line 27: To meet brand standards with these info boxes, we also need to kill this space in between the lines.

website/docs/development/packages/icons/api.md
```diff
@@ -0,0 +1,223 @@
+---
+sidebar_position: 2
+tags: [Development, Usage, Icons]
+---
+
+# Icons
```
caseybaggz, line 6: All the page navigation links should match the file names as close as possible. 😉 

```suggestion
# API
```

website/docs/development/packages/icons/api.md
```diff
@@ -0,0 +1,223 @@
...
+
+To use icons in your application, install the npm package.
+
+```bash npm2yarn
+npm install @pluralsight/icons@alpha
+```
+
+## Usage
+
+:::note
```
caseybaggz, line 24: I visually don't love that we are starting this section off with a giant note in your face. I would prefer we move it to the end of this section.

caseybaggz: I would also change it to a `tip`

website/docs/development/packages/icons/api.md
```diff
@@ -0,0 +1,223 @@
...
+import { menuIcon } from '@pluralsight/icons/svelte'
+```
+
+:::note
+
+If your bundler does not support tree shaking, please see our notes on [minimizing bundle size](#minimizing-bundle-size)
+
+:::
+
+```jsx title="Using an icon element in React"
```
caseybaggz, line 50: So, this section is a lot visually and I think we can clean it up to read better than what we have now (or how MUI does). Likewise, since this is an API page, we need there to be some consistency with what we are doing with HS.

Instead of using a normal codeblock, instead use the LiveExamplesTab and start working on the sandbox contents in a new folder "Icons" that is parrallel to Buttons.

1. Move this to the first item on this section. 
2. Create a subsection (level 3 so it's nested in the side nav but still there) called "Importing by framework" 
3. Move the first note under the LiveExamples component/content.

caseybaggz: For what it's worth, all of our HS methods use the term "props" not "styles" (i.e. `getButtonProps`) since we are delivering props and not actual styles themselves.

website/docs/development/packages/icons/api.md
```diff
@@ -0,0 +1,223 @@
...
+  </span>
+)
+
+export default MenuIcon
+```
+
+## Coloring
+
+By default the icons are monochromatic and will inherit the `color` of their parent.
+
```
caseybaggz, line 66: Kill the space. Add the second sentence to end of first.

website/docs/development/packages/icons/api.md
```diff
@@ -0,0 +1,223 @@
...
+This ensures the icon color matches its context by default and facilitate theming while still allowing for control over the icon's color directly.
+
+```jsx title="Example of icon color inheritance"
+import { menuIcon } from '@pluralsight/icons'
+
+const BlueBlock = (
+  <div style={{ color: '#00f' }}>
```
caseybaggz, line 73: We should be referencing design token values now that it exists and we use them. 😸

website/docs/development/packages/icons/api.md
```diff
@@ -0,0 +1,223 @@
...
+    {/* icon is blue */}
+    <span>{menuIcon}</span>
+    ...
+  </div>
+)
+```
+
+## Sizing
+
+Icons have a 1:1 aspect ratio.
```
caseybaggz, line 84: You can kill this intro line - not really seeing any value here from engineering.

website/docs/development/packages/icons/api.md
```diff
@@ -0,0 +1,223 @@
...
+
+We have defined four standard sizes for icons. This sizing should be applied using the `headless-styles` package.
+
+- `xs`: 16px
```
caseybaggz, line 88: Instead of using a list - please follow the same guide we are using in Buttons via a TS interface block.

caseybaggz: And we need to update these values to use `rem`.

website/docs/development/packages/icons/api.md
```diff
@@ -0,0 +1,223 @@
...
+- `s`: 20px
+- `m`: 24px
+- `l`: 48px
+
+## Accessibility
+
+An accessible label is included by default for each icon in the SVG itself.
```
caseybaggz, line 95: This isn't enough information here, we need to provide an example code block showing what we are doing below it.

website/docs/development/packages/icons/api.md
```diff
@@ -0,0 +1,223 @@
...
+
+## Organization and Naming
+
+We provide `jsx` elements by default. Other formats will be available in corresponding subfolders. e.g., `import { menuIcon } from '@pluralsight/icons/svelte'`
```
caseybaggz, line 99:
```suggestion
We provide two types of elements: React components (default) and SVG elements. Each exported icon is named in camelCase and suffixed with "Icon".
```

website/docs/development/packages/icons/api.md
```diff
@@ -0,0 +1,223 @@
...
+
+Each exported icon is named in camelCase and suffixed with "Icon".
```
caseybaggz, line 101: Kill via change above.

website/docs/development/packages/icons/api.md
```diff
@@ -0,0 +1,223 @@
...
+
+Icons are organized into the same categories as [Material UI icons](https://fonts.google.com/icons).
+
+- Action
```
caseybaggz, line 105: There's no need for this list since you are referencing the source.

website/docs/development/packages/icons/api.md
```diff
@@ -0,0 +1,223 @@
...
+- Navigation
+- Notification
+- Places
+- Search
+- Social
+- Toggle
+
+## Icon Listing
+
+_Coming soon_
```
caseybaggz, line 126: Replace with TBD since we aren't making a movie. 😂

website/docs/development/packages/icons/api.md
```diff
@@ -0,0 +1,223 @@
...
+import menuIcon from '@pluralsight/icons/navigation/menuIcon'
+```
+
+When importing for other frameworks, the framework comes before the category.
+
+```javascript title="Pathed import for Svelte"
+import menuIcon from '@pluralsight/icons/svelte/navigation/menuIcon'
+```
+
+### Babel
```
caseybaggz, line 156: I honestly don't think we need these sections. It's just going to force us to have to keep track of these other sources and docs when it honestly isn't our responsibility. People who use this lib should already know it. We aren't here to help them manage their architecture.

website/docs/development/packages/icons/intro.md
```diff
@@ -11,110 +11,118 @@ This is **unreleased** documentation for Pluralsight Design **icons** package.
...
-### Inline SVG Elements
-
-:::note
+## Pluralsight icons
+
+**TODO**
+
+## Why SVG?
+
+We chose to use inline Scalable Vector Graphics (SVG) after a thorough consideration of alternatives.
```
caseybaggz, line 30: This section needs a complete re-write as it is visually not sexy at all. It needs to be re-written to only be a paragraph that focuses on making the "dicussion" and "rfc" the stars of the paragraph - not the "footnotes".

Definitely kill the overwhelming list of the Benefits which is the exact content the discussion and RFC describe.

website/docs/development/packages/icons/intro.md
```diff
@@ -11,110 +11,118 @@ This is **unreleased** documentation for Pluralsight Design **icons** package.
...
+  - Cons
+    - Cannot control color via CSS
+    - No default accessibility
 
-This library provides only the icon element to allow for maximum flexibility.
-
-Additional attributes will be provided by the `headless-styles` package.
-
-:::
+## Usage
```
caseybaggz, line 78: We need to kill this section since it is used in the API.

website/docs/development/packages/icons/intro.md
```diff
@@ -11,110 +11,118 @@ This is **unreleased** documentation for Pluralsight Design **icons** package.
...
 ```bash npm2yarn
 npm install @pluralsight/icons@alpha
 ```
 
-## Usage
-
-### Inline SVG Elements
-
-:::note
+## Pluralsight icons
```
caseybaggz, line 24: Kill, not sure what you had planned here? 🤷‍♀️

caseybaggz (commented):
OK. There is a lot here but from what I saw - we have the page content backward. It looks like there is way too much content in API which has nothing to do with API, but rather theory (Intro), and the Intro section is left with what appears to be a struggle to come up with content.

So, after you make the changes from the other comments - then reorder the content to contain the following:

## Intro
* SVG vs. Icon font war // this is the new title for the "Why SVG?"
* Accessibility
* Organization & Naming
* Minimizing Bundle Size

## API
* Installation
* Basic Example (all of our API pages should start with Live examples - see Button)
* Icon Colors (live example)
* Icon Sizes (live example)
* Icons // new title for old "Icon Listing"
* Icon Size Guide (actual sizing content you have)
* Importing SVG Files (fallback)

website/docs/development/packages/icons/api.md
Rykus0: Thanks! I wasn't sure if 'API' would be the best title for the page (as you noticed, I need to rearrange things still), but I couldn't reuse 'icons' as the filename, so I went with it.  But I think it works, especially after I've rearranged things.

website/docs/development/packages/icons/api.md
Rykus0: Thanks, I wasn't aware of that.  I'll fix it.

website/docs/development/packages/icons/api.md
Rykus0: 🤣 Yeah, I meant to change that.  Channeling the early 2000s :rofl:

website/docs/development/packages/icons/api.md
Rykus0: Ok, thanks.  I had started with a normal code block for now and was going to update to the live example later, but I can totally jump into that now.  I also thought I pulled`getIconStyles` from somewhere else in the docs but can't find it now.  Maybe I was looking at a really old version.

website/docs/development/packages/icons/api.md
Rykus0: I was thinking of it as a guideline in case someone applies disproportionate custom sizes, but wasn't considering that the svg viewbox handles that.  I'll take it out.

website/docs/development/packages/icons/intro.md
Rykus0: You had suggested this, referencing the MUI docs, but I didn't know what to say about our icons in a similar vein.  I was going to reach out to the designers to see if they had any info that would be good to include here.  I can remove it, and we can always change it if we decide to later.

website/docs/development/packages/icons/api.md
Rykus0: Sorry, I was trying to keep it light (and probably being a bit lazy).  I'll update with token values

PR #147, opened 2022-04-01
Feat/icons foundation

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

None

## What is the new behavior?

Added source icons and script to format. 
- Removes width/height in favor of viewBox
- Adds a class (ps-icon-svg)
- Adds an aria-label if not already present
- Adds role='img' if not present
- Makes svg files available from @pluralsight/icons/svg/*.svg
- Optimizes SVG syntax
- Sets fill to currentColor


caseybaggz:
I think this PR is a bit premature since we are still surveying the docs. Likewise, there is no categorization that we agreed upon within the `svg` directory. I would also like you to move the `utils` into the `src` to keep the directory setup standard. 

Lastly, I would prefer if we keep this PR broken into much smaller chunks by category. It is nearly impossible to rationally review a true 182 file PR (not bin files) which exponentially increases the risk factor of us missing something in the code.

On a total material level, this will also make us look pretty bad in the Flow worklog chart.

PR #160, opened 2022-04-06
feat: make flow tokens

### LINKED ISSUE #157: [Feature?]: allows tokens to use multi-org setup
### Duplicates

### Latest version

### Summary 💡

Flow will need their own theme to support both CSS and CSS-in-JS which means we need to convert tokens to use a mutli-org setup and structure which also impacts the current filters being used.

### Examples 🌈

_No response_

### Motivation 🔦

_No response_

caseybaggz:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

closes #157 

## What is the new behavior?
* Updates tokens to include Flow config & theming
* Updates docs to include Flow specific theme information

## Other information
* Each platform now has a devoted config for token management
* Flow tokens are added using the same design 
  * base directory = private
  * public/flow = what is shipped to user
* Everything is already configured, Flow team just needs to add hsl values to base files and source each theme in the correct public file


website/docs/development/packages/tokens/colors.mdx
```diff
@@ -23,25 +23,37 @@ The `headless-styles` package will consume the `css` and `js` tokens internally.
...
 
 :::caution
 Themes **depend on the normalize.css** file to be used in your project. Check out our [getting started](../../getting-started/installation.md#normalizecss) docs to learn how to add the normalize.css file to your project.
 :::
 
+### Flow theme tokens
+
+For any of our tokens, we highly recommend using the CSS solution as you will have a faster experience for your team or product. By nature, all of our themes are shipped with our normalize.css file.
+
+However, if you prefer to use CSS-in-JS, you can easily import the Flow theme by targeting it's directory in our package.
```
alexzywiak, line 48: Typo it's -> its

packages/design-tokens/tokens/public/flow/color.yaml
```diff
@@ -0,0 +1,123 @@
...
+# for each theme - not create additional unique ones).
+
+# This file should only export the dark theme colors. Other themes
+# will be declared in our normalize.css file. This is to keep performance
+# high and prevent "token vomit" (i.e. we just need to overwrite tokens
+# for each theme - not create additional unique ones).
+
+ps:
+  background:
+    value: '{color.accent.background.dark.value}'
```
alexzywiak, line 15: Do these need to to have the `flow` prefix in there? `color.accent.background.flow-dark.value`

packages/design-tokens/normalize.css
```diff
@@ -147,3 +148,53 @@ html[data-theme='light'],
   --ps-warning-text-weak: hsl(43deg 100% 32% / 100%);
   --ps-warning-text-medium: hsl(43deg 100% 42% / 100%);
 }
+
+html[data-theme='flow-light'],
+.flow-light {
```
alexzywiak, line 153: Trying to wrap my head around the best way to use this.  

The flow-dark css theme vars will be generated into `build/flow/css/variables.css`
To switch to the flow-light theme, we add a class name (or data-theme?)`flow-light`, but are the values in `normalize.css` generated?  How do we update them based on the `.yaml` files?

packages/design-tokens/tokens/public/flow/color.yaml
caseybaggz: @alexzywiak yes! So, you will need to first - add all the values to the "base" empty spots. For consistency, please use `hsl` modern syntax which is what we now use on the design files.

Then, you source those in this file using the **exact** syntax you just wrote down!

packages/design-tokens/normalize.css
caseybaggz: Values in normalize are not generated - you gotta do it manually because some themes share values for light/dark so you don't need to overwrite everything and the token system isn't that smart.

website/docs/development/packages/tokens/colors.mdx
caseybaggz: Done.

packages/design-tokens/tokens/public/flow/color.yaml
caseybaggz: Alrighty, I went ahead and just pushed up this update, so all you guys literally have to do is just add the tokens in the "base" directory files! 🎉

alexzywiak (approved):
Love it.  May not be able to get some UX input on the correct hsl values for a bit, it won't ruin things to have those empty for now, right?

caseybaggz:
> Love it. May not be able to get some UX input on the correct hsl values for a bit, it won't ruin things to have those empty for now, right?

@alexzywiak nope. You just will have an empty string for a value I think? If it helps, all you have to do is select HSL in the Figma dropdown option on the right sidebar thing.

Screen Shot 2022-04-07 at 8 21 53 AM

Also, very very very important that all the colors when used in combination are passing contrast ratios for accessiblity.

Rykus0 (approved):

Issue #154, opened 2022-04-05
[Feature?]: add workspace for CSS transpilation tool


caseybaggz:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

We are currently using https://github.com/americanexpress/css-to-js to create JS style objects from the original CSS files in `headless-styles`. However, there are a few bugs & features we can utilize by customizing the lib to meet our needs.

### Motivation 🔦

Help reduce engineering hours when building HS APIs

PR #164, opened 2022-04-07
feat: generate TS files from CSS

caseybaggz:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

contributes to #154 

## What is the new behavior?
Adds custom plugin `css-to-ts` in `sandbox` which reads all CSS files and converts them to generated .ts files for JS APIs in headless-styles

## Other information
THIS DOES NOT REPLACE CSS VARIABLE NAMES. My next step is to create a theme map to replace the values.


packages/headless-styles/sandbox/plugins/css-to-ts/convert-inputs/write-file.js
```diff
@@ -0,0 +1,58 @@
...
+`
+
+const writeToFile = (outPath, input, outputType) => {
+  let fileOutPath = outPath
+
+  // Case: "SplitFile": Export input object key, values as exports to separate file
+  if (outputType === 'splitFile') {
+    // Create dir for generated split files
+    if (path.extname(fileOutPath)) {
+      fileOutPath = fileOutPath.slice(0, -3)
```
Rykus0, line 19: Is it safe to assume the extension length here?

caseybaggz: @Rykus0 this is from the original lib, so yes.

packages/headless-styles/sandbox/plugins/css-to-ts/convert-inputs/convert-inputs.js
```diff
@@ -0,0 +1,38 @@
+import path from 'path'
+import fs from 'fs'
+import convertToJS from '../convert-rules/convertToJS'
+import reverseMediaQueries from '../utils/mediaReverse'
+
+export const convertStringToJson = (input, mediaReverse) => {
```
Rykus0, line 6: Should we be using regular function notation instead of arrow functions here?

packages/headless-styles/sandbox/plugins/css-to-ts/convert-inputs/write-file.js
```diff
@@ -0,0 +1,58 @@
...
+    } else {
+      fileOutPath = path.join(fileOutPath, input.filename)
+    }
+
+    if (!existsSync(fileOutPath)) {
+      mkdirSync(fileOutPath)
```
Rykus0, line 25: Looks like, in one case we'll have the path with the extensionless filename (path/input.js -> path/input/), and in the other case the path will include the file extension (path/ + input.js -> path/input.js/).  I could be reading it wrong, but if not, is that the desired behavior?

packages/headless-styles/sandbox/plugins/css-to-ts/convert-inputs/write-file.js
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    Object.entries(input.contents).forEach((key) => {
+      const body = indentObject(key[1], 2)
+      writeFileSync(path.join(fileOutPath, `${key[0]}.ts`), 'export default ')
+      appendFileSync(path.join(fileOutPath, `${key[0]}.ts`), body)
```
Rykus0, line 31: I think it would be better to cache the path + filename into a variable here

```suggestion
      const fileOut = path.join(fileOutPath, `${key[0]}.ts`)
      writeFileSync(fileOut, 'export default ')
      appendFileSync(fileOut, body)
```

packages/headless-styles/sandbox/plugins/css-to-ts/convert-inputs/write-file.js
Rykus0: Yeah, I wasn't considering that at first, but I'm looking at it with that in mind, now

packages/headless-styles/sandbox/plugins/css-to-ts/convert-inputs/convert-inputs.js
Rykus0: @caseybaggz - I'm going to close these comments, since they refer to the original plugin, and revising that isn't part of this effort

Rykus0 (approved):

PR #182, opened 2022-04-18
feat(hs): create initial getBadgeProps API

caseybaggz:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

contributes to #180 

## What is the new behavior?
Adds initial `getBadgeProps` API for CSS & Svelte solutions


packages/headless-styles/package.json
```diff
@@ -2,36 +2,47 @@
...
   "description": "Headless styles for Pluralsight.",
-  "main": "build/common/index.js",
-  "module": "build/index.js",
+  "type": "module",
+  "main": "build/index.js",
+  "module": "build/wrapper.mjs",
   "types": "build/index.d.ts",
+  "exports": {
+    "import": "build/wrapper.mjs",
+    "require": "build/index.js"
```
Rykus0, line 11: Does this need to be `.cjs` because of `"type": "module"`?

packages/headless-styles/src/components/Badge/badgeCSS.module.css
```diff
@@ -0,0 +1,28 @@
...
+  padding-inline-end: 4px;
+  padding-inline-start: 4px;
+  vertical-align: middle;
+  white-space: nowrap;
+}
+
+.weak,
+.medium,
+.strong {
+  composes: psBadgeBase;
```
Rykus0, line 18: Should we rely on the css modules syntax, or would it be better to be more compatible with plain css?  I feel like this creates a dependency, but that is a dependency we're okay with?  I want to check, because I don't usually do that unless I have to, but I'm also not as familiar with this approach and want to make sure I'm understanding it.

packages/headless-styles/package.json
caseybaggz: Yup! Please see our doc updates I posted in the release notes. This follows the Node standard down to the T!

packages/headless-styles/src/components/Badge/badgeCSS.module.css
caseybaggz: Honestly, most the world uses CSS modules now so I think it's fine. Also, Svelte doesn't natively support it - so in that instance you see me adding the static class which you can consider the "progressive enhancement" if you will.

Either way, that's one of the topics we will have in the Intro file - that we use CSS modules and people will need to make sure their bundler is setup with said plugin. Most already come with it baked in now. 👍

packages/headless-styles/package.json
caseybaggz: Oh, sorry! I misread that - it is supposed to, but doesn't have to be. The TS team has basically been a bunch of A-holes on the topic saying they aren't responsible for maintainting extensions...when a very very large chunk of the user base has already requested this feature (exactly for this context - .cjs & .mjs).

I'm changing the module name, but don't want to add another script to convert the rest of the files. We'll do that if we have to for this package later since it will just make our build process more insane than TS is forcing us to be (due to the extension debackle).

Rykus0: Ok, yeah, I didn't realize that.  I thought those would be the outputs and not that we'd have to rename.  And I haven't closely read the spec on the `exports` property, so I'll be sure to do that.

Rykus0 (approved):

PR #189, opened 2022-04-21
feat(icons): optimize and format icons via svgo

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [x] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

n/a

## What is the new behavior?

Adds an SVG build command to the `icons` package which:

1. Removes previously generated svg files
3. Runs `svgo` optimize, compress, and format the svg files in the following ways:
  - removes `xmlns` attribute
  - removes explicit dimensions (replacing with `viewBox`, if missing)
  - removes coloring attributes (stroke|fill|fill-opacity|opacity)
  - sets `fill="currentColor"` to enable color inheritance
  - adds `role="img"` 
  - adds `class="ps-icon-svg"`
  - adds `aria-label="${filename} icon"` where `filename` is the file name minus extension and words separated by a space


packages/icons/package.json
```diff
@@ -8,9 +8,14 @@
   "files": [
     "build/*"
   ],
+  "exports": {
```
caseybaggz, line 11: There are a few cleanup things we need to do for this repo. 

1. First and foremost, let's move the devDeps under the scripts, then update the package to use our new Node modue format which you can find in any of the other packages now.
2. I prefer to utilize `pre` scripts so you don't have to duplicate a name just to "clean" things - so let's convert `clean:svg` to instead use `prebuild:svg`, then you can keep the script clean with just the build task.
3. For maintainability and easier readability, please use the longform script flags so people know exactly what is going on (for example, I have no idea what `-f` and `-r` is and I shouldn't have to reference the docs for it.

packages/icons/src/utils/svgo/addAriaLabel.js
```diff
@@ -0,0 +1,63 @@
+'use strict'
```
caseybaggz, line 1: For consistency, let's please use the latest Node standard - so converting this file to use .mjs and update the imports.

This is also manipluating the exports object which is an anti-pattern. We should instead be defining constants and returning that in the `module.exports`. However, this will all go away when you convert to using `.mjs` (but for future reference)

caseybaggz: I also don't think it needs to be as heavily nested in the directory that you have it. It should just live in the root of the `utils` which is pretty standard.

packages/icons/svgo.config.js
```diff
@@ -0,0 +1,39 @@
+const addAriaLabel = require('./src/utils/svgo/addAriaLabel.js')
+
+module.exports = {
```
caseybaggz, line 3: Same here, but by nature - config files should use `.cjs`

packages/icons/package.json
```diff
@@ -2,17 +2,29 @@
...
   "types": "build/index.d.ts",
+  "exports": {
+    "import": "build/wrapper.mjs",
+    "require": "build/index.cjs",
+    "./svg/": "./build/svg/"
+  },
   "files": [
     "build/*"
   ],
+  "sideEffects": false,
```
caseybaggz, line 17: I would actually keep the `index` file listed in here just in case. Webpack even recommends it as a good practice.

packages/icons/src/utils/addAriaLabel.cjs
```diff
@@ -0,0 +1,71 @@
...
+    },
+  }
+}
+
+module.exports = {
+  name,
+  type,
+  active,
+  description,
+  fn: addAriaLabel,
```
caseybaggz, line 70: I think the API would be much cleaner if you just exported and used `addAriaLabel`. Using `fn` is honestly kind of weird and outdated.

Rykus0: `fn` is the property that svgo expects for the plugin, so I have to use it (unless I'm missing something)

packages/icons/package.json
Rykus0: Good to know!  I'll fix that up

caseybaggz (approved):

Issue #195, opened 2022-04-22
[Feature?]: Add source SVG files


Rykus0:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Copy the existing SVG files from the Classic DS.
Rename to remove `.icon`.
Sort into folders based on updated categorization.

### Motivation 🔦

Icon refresh is still a ways out, so we will be proceeding with the existing icons for now.  Thoughtful categorization will help us maintain icons by allowing us to easily target subsets for separate processing if needed.

The source SVG files are required to generate our consumable icon sets - even if that solution changes.

PR #208, opened 2022-04-26
feat(icons): add common-action source icons (part 1)

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

Contributes to #195

## What is the new behavior?

Adds common-action icons from Classic DS

## Other information

1 of 2 chunks of icons for this category.
Added xmlns attribute


caseybaggz (commented):
I'm noticing a pretty big inconsistency with the icon svgs. Some have no inner padding others do. We need to make sure they all are without the inner padding (like the brands).

Rykus0:
@caseybaggz I didn't double-check the first few batches, so I'll go back to update those in a later pass.  I've been checking the `viewBox` to make sure it's square (and plan to add this into the build process), because this was an issue with some of the "alt" brand icons due to the viewbox being rectangular, but I hadn't checked the dimensions. (I'd assumed they were okay since they came from Classic - which is a reminder of why not to make assumptions!)

For the ones that are too large (looking at "cloud" for example) - I see that is wrong in Figma (based on the [design guidelines](https://design-system.pluralsight.com/core/iconography)).   I don't have access to change them there, but I think I can fix it with an adjustment the viewBox

There are some others (a couple of the alts that I adjusted) that look like the issue was actually in the export, so I'll try that and if not, I'll see if I can fix those by adjusting the viewBox as well.

caseybaggz (commented):
So, I would say we went in the opposite direction. We actually want the icons to have *no* padding (which matches the branding icons and gives a more consistent usage for engineering) where as you have added it, which unfortunately is still inconsistent with that.

I would revert and figure out how to remove the access padding on the icons even if it means re-exporting the icon on the specific layer of the svg and not the group.

PR #232, opened 2022-05-02
feat(icons): add svgo plugin to square viewBox attribute

### LINKED ISSUE #223: [Feature?]: Automatically square rectangular viewboxes
### Duplicates

### Latest version

### Summary 💡

Create an `svgo` plugin that will adjust non-square viewboxes to be square.  This is safe if we only expand the smaller dimension, since that will not affect the aspect ratio of the rendered image.  The corresponding offset change can be calculated as half of the difference between the dimensions.  Any existing offsets will also need to be preserved in the final value.

By adding this to the build process we can ensure consistent results with less effort.

(Using an `svgo` plugin is not required, but leverages an existing tool that is already iterating over the files and performing similar actions.  If the tooling were to change in the future, the logic could be easily extracted for use in the new tools.)

### Examples 🌈

```javascript
// assume horizontal rectangle (w > h) for example
const delta = width - height
const newOffset = xOffset - (delta / 2)
const viewBox = [newOffset, yOffset, width, width].join(' ')
```

### Motivation 🔦

The primary goal is to ensure a consistent expectation in baseline positioning of icons, without additional styling or special cases.  

Additionally, we want to reduce the possibility of error as well as reducing the effort of adding new icons.  Since it is easiest to export the icon with the viewBox equal to its dimensions, automating this process would mean one less concern in the creation of the SVGs.

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

closes #223 
Some icons have rectangular viewBoxes, which may cause alignment issues.  

## What is the new behavior?

Rectangular viewBoxes will be expanded to be square such that the image itself is not changed - the aspect ratio and original offsets are preserved.

### DIFF
--- a/packages/icons/utils/squareViewBox.cjs
+++ b/packages/icons/utils/squareViewBox.cjs
@@ -7,21 +7,6 @@ const type = 'visitor'
 const active = true
 const description = 'Adjusts viewbox to be a square if it is not'
 
-/**
- * Adjusts viewBox so that it is a square by expanding the smaller side and
- * adjusting the offset to preserve the original positioning.
- *
- * @example
- * plugins: [
- *   squareViewBox,
- * ]
- *
- * <svg viewBox="0 0 24 12">...
- *             ⬇
- * <svg viewBox="0 -6 24 24">...
- *
- * @author Tom Pietrosanti
- */
 function squareViewBox() {
   return {
     element: {
--- a/packages/icons/utils/squareViewBox.cjs
+++ b/packages/icons/utils/squareViewBox.cjs
@@ -1,0 +1,62 @@
+'use strict'
+
+const path = require('path')
+
+const name = 'squareViewBox'
+const type = 'visitor'
+const active = true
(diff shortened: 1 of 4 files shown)


packages/icons/utils/squareViewBox.cjs
```diff
@@ -0,0 +1,62 @@
...
+
+const path = require('path')
+
+const name = 'squareViewBox'
+const type = 'visitor'
+const active = true
+const description = 'Adjusts viewbox to be a square if it is not'
+
+/**
+ * Adjusts viewBox so that it is a square by expanding the smaller side and
```
caseybaggz, line 11: So, as a company, we prefer to follow the "Clean Code" standards of keeping files free from comments unless it is absolutely necessary - which this isn't. It's just a giant block of code describing what your code does which is easy to read because you are using really good names.

So, let's kill this and take it to heart for future files. 😉

caseybaggz (approved):

Issue #220, opened 2022-04-29
[Feature?]: add getCirclularProgressProps & getJSCircularProgressProps


caseybaggz:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

- [x] CSS API
- [x] JS API
- [x] Chakra API
- [x] SC sandbox
- [x] Svelte sanbox
- [x] MUI sandbox
- [x] Chakra sandbox

caseybaggz:
Everything pushed on the latest. MUI is incompatible, Chakra has a bug that we filed and are watching for the fix. Clsoing.

PR #237, opened 2022-05-03
feat: add CircularProgress CSS API

caseybaggz:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

contributes #220 

## What is the new behavior?
Adds CircularProgress CSS API


packages/headless-styles/sandbox/src/components/CircularProgress.jsx
```diff
@@ -0,0 +1,31 @@
...
+  return (
+    <div {...determinateProps.containerProps}>
+      <svg {...determinateProps.svgBoxProps}>
+        <circle {...determinateProps.baseCircleProps} />
+        <circle {...determinateProps.nowCircleProps} />
+      </svg>
+      {props.showValue && (
+        <span {...determinateProps.labelProps}>
+          {determinateProps.labelProps.value}
+        </span>
```
Rykus0, line 15: I read that `role="progressbar"` will cause all descendants to be presentational, so this label won't be read to screen readers.  We could add an `aria-label` to the `a11yProps` to account for that.

caseybaggz: Right, but you don't need the label because you have the a11y props already - so it's redundant. This is why it's not used as a common pattern for `progressbar`.

Rykus0: Ah, right.

packages/headless-styles/src/components/CircularProgress/shared.ts
```diff
@@ -0,0 +1,47 @@
...
+export function getA11yCircularProgressProps(
+  a11yOptions?: CircularProgressA11yOptions
+) {
+  return {
+    [a11yPropMap.valueMax]: a11yOptions?.max,
+    [a11yPropMap.valueMin]: a11yOptions?.min,
+    [a11yPropMap.valueNow]: a11yOptions?.now,
+    role: a11yRole,
+  }
+}
```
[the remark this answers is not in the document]
caseybaggz: Done.

PR #260, opened 2022-05-16
Feat/hs get js icon props

### LINKED ISSUE #252: [Feature?]: Add getJSIconProps API
### Duplicates

### Latest version

### Summary 💡

Add the css-in-js API for getIconProps

### Examples 🌈

_No response_

### Motivation 🔦

_No response_

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

closes #252 

## What is the new behavior?

Adds CSS in JS API for Icon

## Other information

Changes aria-hidden output to return a string instead of boolean
Renames default prop getter to align with conventions (`getDefaultOptions` -> `getDefaultIconOptions`)

### DIFF
--- a/packages/headless-styles/sandbox/src/components/Icon.jsx
+++ b/packages/headless-styles/sandbox/src/components/Icon.jsx
@@ -1,0 +1,43 @@
+import { getIconProps, getJSIconProps } from '../../../src'
+
+function PSIcon(props) {
+  return (
+    <svg
+      aria-label="pluralsight logo"
+      viewBox="0 0 20 20"
+      role="img"
+      fill="currentColor"
+      className="ps-icon-svg"
+      {...props}
+    >
+      <path
+        fillRule="evenodd"
+        clipRule="evenodd"
+        d="M7.5 7.203 5.312 5.937v8.125L7.5 12.797v2.437L16.473 10 7.5 4.766v2.437Zm0 1.083-1.25-.723v4.874l1.25-.723V8.286Zm.938 2.886V8.828L10.463 10l-2.025 1.172Zm0 1.083L12.335 10 8.437 7.745V6.398L14.614 10l-6.175 3.602v-1.347Z"
+      />
+      <path
+        fillRule="evenodd"
+        clipRule="evenodd"
+        d="M10 20c5.523 0 10-4.477 10-10S15.523 0 10 0 0 4.477 0 10s4.477 10 10 10Zm0-.938A9.062 9.062 0 1 0 10 .939a9.062 9.062 0 0 0 0 18.125Z"
+      />
+    </svg>
+  )
+}
+
+export default function Icon(props) {
+  if (props.logJS) {
+    console.log({ ...getJSIconProps({ size: 'xs' }) })
+  }
+
+  return (
+    <div id="icon">
+      <h3>Icon</h3>
+      <div className="App-container">
+        <PSIcon {...getIconProps({ label: 'extra small', size: 'xs' })} />
+        <PSIcon {...getIconProps({ label: 'small', size: 's' })} />
+        <PSIcon {...getIconProps({ label: 'medium (default)', size: 'm' })} />
+        <PSIcon {...getIconProps({ label: 'large', size: 'l' })} />
+      </div>
+    </div>
+  )
+}
--- a/packages/headless-styles/src/components/Icon/iconJS.ts
+++ b/packages/headless-styles/src/components/Icon/iconJS.ts
@@ -1,0 +1,31 @@
+import { createJSProps, transformStyles } from '../../utils/helpers'
+import { getA11yIconProps, getDefaultIconOptions } from './shared'
+import styles from './generated/iconCSS.module'
+import type { IconOptions } from './types'
+
+export const ChakraIcon = {
+  baseStyle: styles.psIcon,
+  defaultProps: {
+    size: 'm',
+  },
+  sizes: {
+    xs: styles.xsIconSize,
+    s: styles.sIconSize,
+    m: styles.mIconSize,
+    l: styles.lIconSize,
+  },
+}
+
+export type StyleKey = keyof typeof styles
+
+export function getJSIconProps(options?: IconOptions) {
+  const { size, ...a11y } = getDefaultIconOptions(options)
+  const a11yProps = getA11yIconProps(a11y)
+  const sizeKey = `${size}IconSize`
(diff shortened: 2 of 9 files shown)


packages/headless-styles/sandbox/src/components/Icon.jsx
```diff
@@ -0,0 +1,43 @@
+import { getIconProps, getJSIconProps } from '../../../src'
+
+function PSIcon(props) {
+  return (
+    <svg
```
caseybaggz, line 5: We should be using an Icon from our library here not raw markup.

packages/headless-styles/src/components/Icon/iconJS.ts
```diff
@@ -0,0 +1,31 @@
...
+export type StyleKey = keyof typeof styles
+
+export function getJSIconProps(options?: IconOptions) {
+  const { size, ...a11y } = getDefaultIconOptions(options)
+  const a11yProps = getA11yIconProps(a11y)
+  const sizeKey = `${size}IconSize`
+  const jsStyles = styles[sizeKey as StyleKey]
+
+  return {
+    a11yProps,
```
caseybaggz, line 28: Since `a11yProps` is not being reused, I would just use the function here to keep the API code cleaner and have fewer lines.

packages/headless-styles/src/components/Icon/shared.ts
```diff
@@ -23,7 +23,7 @@ export function getDefaultOptions(options?: IconOptions) {
 
 export function getA11yIconProps(a11yOptions?: IconA11yOptions) {
   return {
-    [a11yPropMap.ariaHidden]: a11yOptions?.ariaHidden,
+    [a11yPropMap.ariaHidden]: a11yOptions?.ariaHidden ? 'true' : 'false',
```
caseybaggz, line 26: I would clean this up a little and require a String of `'true' | 'false'` in the API that has a default value of `'false'` so all you have to do is this:

```
[a11yPropMap.ariaHidden]: a11yOptions.ariaHidden
```

caseybaggz: The less logic, the more dependable our API is. 👍

packages/headless-styles/tests/icon/iconJS.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    role: 'img',
+  }
+
+  test('should return a distinct height and width difference based on the size', () => {
+    const xsIconProps = getJSIconProps({ size: 'xs' })
+    const mIconProps = getJSIconProps({ size: 'm' })
+    const sIconProps = getJSIconProps({ size: 's' })
+    const lIconProps = getJSIconProps({ size: 'l' })
+
+    expect(xsIconProps.styles.height).not.toEqual(mIconProps.styles.height)
```
caseybaggz, line 15: It's considered a code smell to use negation testing or comparisons. All these tests should instead use `toEqual` and not `.not.toEqual`.

I would also say this test is doing too much. Instead, break it up into an individual test per size so that way you know if a specific size fails vs this execution will fail and you will have to go out of your way to figure out which part of the test failed.

caseybaggz: Additionally, you never want to use dynamic values in unit tests. You can for the starting value, but all the comparision values (ending) should be static as per the guidelines from the Cypress team (pretty much the ultimate authority on solid testing these days).

packages/headless-styles/sandbox/src/components/Icon.jsx
Rykus0: No problem.  I was going to at first, but then I thought it might be good to demonstrate that the headless styles don't need to go with our library.

packages/headless-styles/src/components/Icon/shared.ts
Rykus0: I was thinking boolean was easier to work with/less chance of error as an input from the API side, so I went that way.  The string definitely simplifies the internals, and since it's not that much of a concern (definitely less with TS), then I'll change it.

packages/headless-styles/tests/icon/iconJS.test.ts
Rykus0: Yeah, I thought it might be doing too much.  I'll get this cleaned up.

Rykus0: Also, just added their best practices guide to my reading list ;)

packages/headless-styles/tests/icon/iconJS.test.ts
```diff
@@ -6,26 +6,56 @@ describe('icon JS', () => {
     role: 'img',
   }
 
-  test('should return a distinct height and width difference based on the size', () => {
+  test('should accept an "xs" size', () => {
     const xsIconProps = getJSIconProps({ size: 'xs' })
-    const mIconProps = getJSIconProps({ size: 'm' })
+
+    expect(xsIconProps.cssProps).not.toHaveLength(0)
```
caseybaggz, line 12: I found that we can be more definitive with the JS tests by specifically targeting styles you know should be different vs. what is being done here.

So for example, something like:
```
expect(xsIconProps.styleProps.width).toEqual('1rem')
```

website/docs/development/packages/headless-styles/Icon.mdx
```diff
@@ -142,7 +142,7 @@ export interface IconA11yOptions {
 interface IconProps {
   className: string
   'aria-label'?: string
```
caseybaggz, line 144: I would remove the optional syntax here and just keep it on the A11yOptions - much easier to read.

packages/headless-styles/tests/icon/iconJS.test.ts
```diff
@@ -1,61 +1,40 @@
...
-    expect(mIconProps.styles).toBeDefined()
-  })
-
-  test('should accept an "l" size', () => {
-    const lIconProps = getJSIconProps({ size: 'l' })
-
-    expect(lIconProps.cssProps).not.toHaveLength(0)
-    expect(lIconProps.styles).toBeDefined()
-  })
+  function testDimensions(size: Size) {
```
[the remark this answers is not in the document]
Rykus0: Ok, interesting.  I'll switch it over.  For what it's worth, my thinking was, since we're testing different values of the same properties, it made sense for a more consistent test case.

Issue #244, opened 2022-05-06
[Feature?]: add getSwitchProps & getJSSwitchProps API


caseybaggz:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

- [x] CSS API
- [x] JS API
- [x] SC SBox
- [x] Chakra SBox
- [x] MUI SBox
- [x] Svelte SBox

caseybaggz:
All sandboxes created. Closing.

PR #261, opened 2022-05-16
Feat/244 switch css api

caseybaggz:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

contributes to #244 

## What is the new behavior?
Adds CSS API for Switch


packages/headless-styles/src/components/Switch/switchCSS.ts
```diff
@@ -0,0 +1,82 @@
...
+const SWITCH = 'ps-switch'
+
+export function getSwitchProps(options?: SwitchOptions) {
+  const { htmlFor, size, value, ...defaultOptions } =
+    getDefaultSwitchOptions(options)
+  const labelClass = `${size}Label`
+  const trackClass = `${size}Track`
+  const thumbClass = `${size}Thumb`
+  const a11yInputProps = {
+    'aria-disabled': defaultOptions.disabled,
```
Rykus0, line 15: Are we using `aria-disabled` instead of `disabled` because we can't guarantee they use an `input` element? (If so, we'll need to add a `role` and maybe some others)

packages/headless-styles/src/components/Switch/types.ts
```diff
@@ -0,0 +1,18 @@
+export interface SwitchA11yOptions {
+  htmlFor: string
+  disabled?: A11yBoolean
+  invalid?: A11yBoolean
+  value: A11yBoolean
```
Rykus0, line 5: Do you think the name `value` might cause confusion with the way the value works for an HTML checkbox?  It would for me, but I'm not the target audience :)

packages/headless-styles/src/components/Switch/switchCSS.ts
caseybaggz: This is another case of overanalyzing I think? This is the standard pattern for meeting a11y guidelines and what other UI libs are doing. Likewise, it's not our job to give unnecessary props just to appease people who purposefully choose not to follow our documentation.

If you are a user who is choosing to purposefully not use an input even though we are showing to use it and it is an a11y standard - then you shouldn't be catered to IMHO.

packages/headless-styles/src/components/Switch/switchCSS.ts
```diff
@@ -0,0 +1,82 @@
...
+        ...createSvelteObj(`${SWITCH}-thumb thumb ${size}Thumb`),
+      },
+      wrapper: {
+        ...a11yRole,
+        ...createSvelteObj(`${SWITCH} base`),
+      },
+    }
+  }
+
+  return {
```
Rykus0, line 58: I could be missing something, but I think we can do this with fewer elements, which would be simpler.  I've used only a container, input, and label before.  But I could be missing some context or over-simplifying.

packages/headless-styles/src/components/Switch/types.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export interface SwitchOptions extends SwitchA11yOptions {
+  label?: string
+  size?: Size
+  tech?: Tech
+}
+
+// types
+
+export type A11yBoolean = 'true' | 'false'
```
Rykus0, line 16: I called mine `BooleanString` - want me to change it to be consistent?

packages/headless-styles/src/components/Switch/switchCSS.ts
```diff
@@ -0,0 +1,82 @@
...
+    type: 'checkbox',
+  }
+  const a11yHidden = {
+    'aria-hidden': 'true',
+  }
+  const a11yRole = {
+    role: 'group',
+  }
+  const inputValues = {
+    'data-checked': value,
```
Rykus0, line 27: If we use `aria-checked` instead, it would get picked up by screen readers too.  I think we'd also need to add a role to it (either "switch" or "checkbox"), especially if we're hiding the `input` element.

packages/headless-styles/src/components/Switch/shared.ts
```diff
@@ -0,0 +1,23 @@
+import type { A11yBoolean, SwitchOptions, Size, Tech } from './types'
+
+const defaultSwitchOptions = {
+  disabled: 'false' as A11yBoolean,
+  htmlFor: '',
```
Rykus0, line 5: Should we provide a unique ID as a default, to make sure it doesn't break if omitted?

packages/headless-styles/src/components/Switch/switchCSS.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  if (defaultOptions.tech === 'svelte') {
+    return {
+      input: {
+        ...a11yInputProps,
+        ...createSvelteObj(`${SWITCH}-input input`),
+      },
+      label: {
+        htmlFor,
```
Rykus0, line 38: If we need to convert `className` to `class` for svelte, do we also need to convert `htmlFor` to `for`?

packages/headless-styles/src/components/Switch/switchCSS.module.css
```diff
@@ -0,0 +1,112 @@
...
+  font-size: 1rem;
+  font-variation-settings: "'wght' 400";
+  font-weight: 400;
+  margin-bottom: 0;
+  margin-inline-end: 0.625rem;
+  opacity: 1;
+  text-align: start;
+  transition-duration: 200ms;
+  transition-property: background-color, border-color, color, fill, stroke,
+    opacity, box-shadow, transform;
```
Rykus0, line 40: Is the transition for switching themes? (or similar)

packages/headless-styles/src/components/Switch/switchCSS.ts
```diff
@@ -0,0 +1,82 @@
...
+        ...createSvelteObj(`${SWITCH}-label label ${size}Label`),
+      },
+      switchContainer: createSvelteObj(`${SWITCH}-container container`),
+      switchTrack: {
+        ...a11yHidden,
+        ...inputValues,
+        ...createSvelteObj(`${SWITCH}-track track ${size}Track`),
+      },
+      switchThumb: {
```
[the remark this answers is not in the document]
caseybaggz: Whoops! Totally forgot to add that style! Good catch!

packages/headless-styles/src/components/Switch/switchCSS.ts
caseybaggz: This is using the same pattern as Chakra to so I don't want to stray too far from it just for consistency since one of our teams is using it. Just helps the styling for our Chakra API a little easier.

packages/headless-styles/src/components/Switch/types.ts
caseybaggz: Sure.

packages/headless-styles/src/components/Switch/switchCSS.ts
caseybaggz: Chakra doesn't do this and they are a11y forward, so I'm not sure we need this.

packages/headless-styles/src/components/Switch/shared.ts
caseybaggz: I don't think so. If someone leaves out this prop, they need to learn why it is breaking. There's a fine line between hand-feeding engineers to the point we are encouraging bad practices (i.e. not knowing to use a `for` attr) and just making a simple mistake.

packages/headless-styles/src/components/Switch/switchCSS.module.css
caseybaggz: What do you mean by switching "themes" - this is a basic property so the CSS knows what other properties to control in the transition animation.

packages/headless-styles/src/components/Switch/switchCSS.ts
caseybaggz: This property is also required so anyone using TS will be forced to use it (our whole team and the majority of the FE world these days).

packages/headless-styles/src/components/Switch/switchCSS.module.css
Rykus0: I meant something outside of the scope of the component, since I didn't see us changing these for the label (maybe I missed it).  So in that case, we're adding these in case whoever is using this component changes those properties for some other reason... a user changing theme was the first example that came to my mind.

packages/headless-styles/src/components/Switch/switchCSS.ts
caseybaggz: Not required, but I changed it. 👍

packages/headless-styles/src/components/Switch/switchCSS.ts
Rykus0: Ok, thanks.  I need to spend some more time with these frameworks to get familiar with what they're doing, how, and why.

packages/headless-styles/src/components/Switch/types.ts
```diff
@@ -0,0 +1,18 @@
+export interface SwitchA11yOptions {
+  htmlFor: string
```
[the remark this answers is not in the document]
Rykus0: Ok, makes sense.

packages/headless-styles/src/components/Switch/switchCSS.module.css
```diff
@@ -0,0 +1,118 @@
...
+/* element states */
+
+.input:focus + .track {
+  box-shadow: 0 0 0 3px var(--ps-background-active);
+}
+
+/* sizes must be last */
+
+.thumb[data-checked='true'] {
+  transform: translateX(calc(3rem - 1.5rem));
```
Rykus0, line 76: Since we can calculate this ahead of time, should we just use the resulting value?

packages/headless-styles/src/components/Switch/switchCSS.module.css
```diff
@@ -0,0 +1,118 @@
...
+}
+
+.sThumb {
+  composes: thumb;
+  height: 0.75rem;
+  width: 0.75rem;
+}
+
+.sThumb[data-checked='true'] {
+  transform: translateX(calc(2rem - 1rem));
```
Rykus0, line 113: Same here - should this just be `translateX(1rem)`?

packages/headless-styles/src/components/Switch/switchCSS.module.css
caseybaggz: You know...I think I'll use some local vars to cut the code down so this is more meaningful.

Rykus0: I see.  That makes sense.  I think it's a good example of where we might benefit from a pre-processor. Since we don't need to calculate those at runtime, we could avoid any performance hits from calc and vars.

Rykus0 (approved):

PR #290, opened 2022-05-23
Docs/icon gallery

### LINKED ISSUE #197: [Docs?]: Display available icons in docs
### Latest version

### Summary 💡

Icon documentation should include a gallery of available icons, sorted into categories.  This listing should be generated automatically from available icons.

### Motivation 🔦

We want to make sure developers can browse the available icons and their categorizations to help quickly find the ones they need.

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

closes #197 

## What is the new behavior?

Adds a page that lists out icons alphabetically by category, with their import name and example.


website/src/components/Icons/IconGallery.js
```diff
@@ -0,0 +1,27 @@
...
+import { getIconProps } from '@pluralsight/headless-styles'
+import styles from './IconGallery.module.css'
+
+function IconGallery(props) {
+  const { icons, size = 'l' } = props
+  const psIconProps = getIconProps({ size: size })
+
+  return (
+    <div className={styles.gallery}>
+      {Object.keys(icons)
```
caseybaggz, line 11: This is a anti-pattern in React. Yes, you can do it - but whenever you are adding logic within a component that has a different responsibility than the parent - it should be it's own component. This also contributes to better performance due to React managing the VD better.

So there are actually 3 components here vs a giant parent doing everything:

* IconGallery - display IconList and pass props
* IconList - loop through icons and display Icon
* IconListItem - display icon content based on `key` being passed in as a prop

Then, because you are passing in a complex list - you should wrap the export of IconGallery in a `memo` function.

caseybaggz (commented):
Sidebar nav links were removed in this process. Should add back.

website/src/components/Icons/IconGallery.module.css
```diff
@@ -8,16 +8,12 @@ html[data-theme='light'] .gallery {
...
-  margin-top: 5rem;
-}
-
 .list {
   display: grid;
   grid-gap: 3rem;
   grid-template-columns: repeat(auto-fit, minmax(12rem, 1fr));
   list-style: none;
-  margin: 2rem 0 0;
+  margin: 2rem 0 4rem;
```
caseybaggz, line 16: This is weird that it didn't throw an error for using shorthand CSS? We'll merge, but let's avoid using it in the future - another CSS anti-pattern for maintainability.

caseybaggz (approved):

website/src/components/Icons/IconGallery.module.css
Rykus0: That makes sense to me in terms of not overriding a setting for one side or another.   In this case I was intentionally setting all 4 sides.  Would we still avoid the shorthand in that case?  I'll do some reading up on it.

Issue #285, opened 2022-05-20
[Feature?]: add Radio API


caseybaggz:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

- [x] CSS API
- [x] JS API
- [x] SC SB
- [x] Chak SB
- [x] MUI SB
- [x] Svelte SB

PR #295, opened 2022-05-24
Feat/285 radio css api

caseybaggz:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

contributes to #285 

## What is the new behavior?
* Adds Radio CSS API
* Updates FormControl API to add new `groupType` option


packages/headless-styles/sandbox/src/components/Radio.jsx
```diff
@@ -0,0 +1,150 @@
...
+            checked={email === '1'}
+          />
+          <RadioInput
+            htmlFor="email"
+            value="2"
+            id="email-2"
+            label="Other"
+            onClick={handleClick}
+            name="email2"
+            checked={email === '2'}
```
Rykus0, line 139: I'm not sure how much it matters for the sandbox example, but `htmlFor` isn't matching `id` in these two.  Also, if all 3 have the same `name` it would work better (e.g., using arrows to select) - but again, I don't know if that matters here.

caseybaggz: Honestly, sandboxes are just that. I'm just trying to test out different cases and see how they work - it's not meant to force any strict rules - just be a dev environment to ensure our styles work and the APIs are easy to use in different use cases.

Rykus0 (approved):

Issue #172, opened 2022-04-12
[Feature?]: add getIconButtonProps API


caseybaggz:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

**DEPENDANT ON ICONS & ICON API BEING DONE**

The headless Button API needs to include:

- [x] CSS API
- [ ] JS API
- [ ] Svelte SB
- [ ] Chakra SB
- [ ] SC SB
- [ ] MUI SB

PR #296, opened 2022-05-24
feat(headless-styles): add getButtonIconLabelProps

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

contributes to #172 

## What is the new behavior?

Adds `getButtonIconLabelProps` which returns button and icon props for when an icon is included as part of the button label


packages/headless-styles/src/components/Button/buttonCSS.ts
```diff
@@ -1,9 +1,21 @@
...
-import type { ButtonOptions, ButtonType, DangerOptions } from './types'
+import type {
+  ButtonOptions,
+  ButtonType,
+  DangerOptions,
+  IconLabelOptions,
+  Size,
+} from './types'
 import styles from './buttonCSS.module.css'
+import { getIconProps } from '../Icon/iconCSS'
```
caseybaggz, line 15: We don't ever want to use HS helpers within others because it will cause too much dependency on each other. Instead - we need our API's to be more compound and allow the user to provide that additional helper if they want/need too.

Let's figure out how to achieve this same thing without the internal helper which conflicts the responsibility of the button API (to only use Button helpers - nothing else).

caseybaggz (commented):
Hmmmm. I don't love this API - there has to be a better way to do this where the icon API controls the icon styling and the button API controls the button styling - not the button doing both. We need this to be much more compound and less complex and tight-nit to each other.

Rykus0:
I'm going to close this one and open a new one when ready with commit comments updated and a more appropriate branch name

PR #303, opened 2022-05-26
Feat/get icon button props

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

contributes to #172 

## What is the new behavior?
Adds `getIconButtonProps` API


packages/headless-styles/src/components/Button/buttonCSS.ts
```diff
@@ -43,3 +54,31 @@ export function getButtonProps(options?: ButtonOptions) {
...
+const iconButtonSizeMap: Record<Size, IconSize> = {
+  xs: 's',
+  s: 'm',
+  m: 'm',
+  l: 'l',
+}
+
+export function getIconButtonProps(options?: IconButtonOptions) {
+  const defaultOptions = getDefaultIconButtonOptions(options)
+  const { kind, size, round, ariaLabel } = defaultOptions
```
caseybaggz, line 67: We don't ship svelte styles with the default api, so you need to break this up into it's own condition. This also assumes forces all of our React teams to have to create custom class props because it's using `defaultClass` and not `className`. Please be sure to follow the current pattern for consistency for v1.

We can then survey to see if the people would rather have something like this in the future. 👍

packages/headless-styles/src/components/Button/types.ts
```diff
@@ -14,6 +14,11 @@ export interface DangerOptions extends ButtonOptions {
   kind?: DangerKind
 }
 
+export interface IconButtonOptions extends ButtonOptions {
+  ariaLabel: string
+  round?: boolean
```
caseybaggz, line 19: I wouldn't use a boolean here for the round property because it's not scalable and will create a massive breaking change if the DS ever changes in the future. Thus, it would be better/safter/scalable to use this:

```
variant: 'default' | 'round'
```

packages/headless-styles/tests/button/buttonCSS.test.ts
```diff
@@ -100,4 +104,119 @@ describe('Button CSS', () => {
       })
     })
   })
+  describe('getIconButtonProps', () => {
```
caseybaggz, line 107: I think we should probably just put these in it's own file. I thought this pattern would scale - but...it's not. 🙈

packages/headless-styles/src/components/Button/types.ts
Rykus0: Good point.  I even had it that way to start and talked myself out of it 🙈

packages/headless-styles/src/components/Button/buttonCSS.ts
Rykus0: hm.  actually I think I'll keep the method with a few changes.  It's still helping to keep things simpler overall

packages/headless-styles/sandbox/src/components/IconButton.jsx
```diff
@@ -0,0 +1,75 @@
+import { getIconButtonProps, getIconProps } from '../../../src'
+import { CodeIcon } from '@pluralsight/icons'
+
+function PSIconButton(props) {
+  const { button, iconOptions } = getIconButtonProps(props)
+
+  return (
+    <button {...button}>
+      <CodeIcon {...getIconProps(iconOptions)} />
```
caseybaggz, line 9: In the future - I wouldn't make the icon static because it only validates how a single icon will look. I would instead just use `props.children` and pass those options to it so you can test out all different icons to make sure they look consistent.

packages/headless-styles/src/components/Button/iconButtonCSS.ts
```diff
@@ -0,0 +1,48 @@
+import { getDefaultIconButtonOptions } from './shared'
+import { createSvelteObj, createCSSObj } from '../../utils/helpers'
+import type { ButtonType, IconButtonOptions, Size, Tech } from './types'
+import styles from './buttonCSS.module.css'
+import type { IconSize } from '../types'
```
caseybaggz, line 5: We should re-order these imports. It's a code-smell to not use a hierarchy teir. There is an eslint rule I need to add to enforce this. It should go...

1. NPM packages
2. Furthest to closest relative locations.

packages/headless-styles/src/components/Button/iconButtonCSS.ts
```diff
@@ -0,0 +1,48 @@
...
+
+interface ButtonClass {
+  defaultClass: string
+  svelteClass: string
+}
+
+function createClass(tech: Tech, classes: ButtonClass) {
```
caseybaggz, line 12: SOB...I've been overthinking this so much. I knew there was a way to do this! 😹  Now you are going to have to make an issue to refactor all the other components to use this pattern. 😛 

Please go ahead and move this function to the global `helpers`, write a test around it, then import that in here.

packages/headless-styles/src/components/Button/shared.ts
```diff
@@ -34,3 +36,21 @@ export function getDefaultDangerOptions(options?: DangerOptions) {
     tech: options?.tech ?? defaultDangerButtonOptions.tech,
   }
 }
+
+export const defaultIconButtonOptions = {
+  kind: 'text' as Kind,
+  size: 'm' as Size,
+  tech: '' as Tech,
+  variant: 'default' as IconButtonVariant,
```
caseybaggz, line 44: I think you can just rename this to `Variant` since it's code-split. It will match the pattern we are already using and be much less to type for our users.

packages/headless-styles/sandbox/src/components/IconButton.jsx
Rykus0: Yeah, I thought of that, but took the easy way out for the sandbox 🙈

PR #326, opened 2022-06-06
Refactor/hs icon a11y

### LINKED ISSUE #304: [Feature?]: HS Icon a11y options
### Duplicates

### Latest version

### Summary 💡

Instead of `label`, explicitly name the argument `ariaLabel` to help educate about accessibility on the web.
Replace `A11yBoolean` with actual boolean in arguments for improved consistency and compatibility.

### Examples 🌈

```typescript
export interface IconA11yOptions {
  ariaLabel?: string
  ariaHidden?: boolean
}
```

### Motivation 🔦

Naming the `label` as `ariaLabel` will increase awareness and arouse curiosity in developers, leading to a higher level of accessibility education.

Using a boolean as an argument is simpler and more efficient than a string - though the final output still needs to be represented as a string.  It is also consistent with how we are handling these types of attributes in other headless-styles APIs

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

closes #304 

API argument that maps to aria-label was named `label`
Used a custom type `A11yBoolean` for aria-hidden

## What is the new behavior?

API argument now named `ariaLabel`
A11yBoolean type removed and replaced with boolean primitive

### DIFF
--- a/packages/headless-styles/src/components/Icon/shared.ts
+++ b/packages/headless-styles/src/components/Icon/shared.ts
@@ -1,4 +1,4 @@
-import type { IconOptions, IconA11yOptions, Tech, A11yBoolean } from './types'
+import type { IconOptions, IconA11yOptions, Tech } from './types'
 import { IconSize } from '../types'
 
 const a11yRole = 'img'
@@ -10,7 +10,7 @@ const a11yPropMap = {
 export const defaultIconOptions = {
   size: 'm' as IconSize,
   tech: '' as Tech,
-  ariaHidden: 'false' as A11yBoolean,
+  ariaHidden: false,
 }
 
 export function getDefaultIconOptions(options?: IconOptions) {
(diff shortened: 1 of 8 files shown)


packages/headless-styles/src/components/Icon/shared.ts
```diff
@@ -1,31 +1,33 @@
-import type { IconOptions, IconA11yOptions, Tech, A11yBoolean } from './types'
+import type { IconOptions, IconA11yOptions, Tech } from './types'
 import { IconSize } from '../types'
 
 const a11yRole = 'img'
 const a11yPropMap = {
```
caseybaggz, line 5: If we are only using these 2 `const` once in the code, we need to kill. For performance reasons, we should only be caching things if they are using 2 or more times. Some rules even go as far as 3. I'm fine with 2 though.

Issue #312, opened 2022-05-27
[Feature?]: Implement class helper in existing APIs


Rykus0:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Replace calls to `createCSSObj` and `createSvelteObj` with singular `createClassProp`

### Examples 🌈

Before
```typescript
  if (tech === 'svelte') {
    return {
      ...a11yProps,
      ...createSvelteObj(`${ICON} psIcon ${sizeClass}`),
    }
  }

  return {
    ...createCSSObj(`${ICON} ${styles[sizeClass]}`, a11yProps),
  }
```

After
```typescript
  return {
    ...a11yProps,
    ...createClassProp(tech, { 
      defaultClass: `${ICON} ${styles[sizeClass]}`,
      svelteClass: `${ICON} psIcon ${sizeClass}`
    }),
  }
```

PR #333, opened 2022-06-07
refactor(headless-styles): use createClassProp for Button

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

Contributes to #312 

## What is the new behavior?

Refactors Button prop getter to use getClassProp.
Also returns `type="button"` for Svelte, which was previously missing

### DIFF
--- a/packages/headless-styles/src/components/Button/buttonCSS.ts
+++ b/packages/headless-styles/src/components/Button/buttonCSS.ts
@@ -1,14 +1,11 @@
 import type { ClassOptions } from '../../utils/helpers'
 import { createClassProp } from '../../utils/helpers'
+import type { Tech } from '../types'
 import { getDefaultOptions, getDefaultDangerOptions } from './shared'
 import type { ButtonOptions, ButtonType, DangerOptions } from './types'
 import styles from './buttonCSS.module.css'
 
-type AllButtonOptions = ButtonOptions | DangerOptions
-
-function createButton(options: AllButtonOptions, classes: ClassOptions) {
-  const tech = options?.tech ?? ''
-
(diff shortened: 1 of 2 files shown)


packages/headless-styles/src/components/Button/buttonCSS.ts
```diff
@@ -1,22 +1,16 @@
...
-  defaultClass: string
-  svelteClass: string
-}
-
-function createButton(options: AllButtonOptions, classes: ButtonClass) {
-  if (options.tech === 'svelte') {
-    return createSvelteObj(classes.svelteClass)
-  }
+function createButton(options: AllButtonOptions, classes: ClassOptions) {
+  const tech = options?.tech ?? ''
```
caseybaggz, line 10: This should already be done in the default props which means it's duplicated logic here. Kill.

caseybaggz: If TS is throwing an error, that means the types need to be correct (AllButtonOptions) vs adding more unnecessary logic to the function.

Rykus0: Yeah, TS didn't like it because `options.tech` could be undefined here.  I think because in the others, we're using the default prop getter right before, but button has the separate method.  I approached it differently last time, making `tech` an optional argument to the helper, which had a bigger impact (need to reorder the args), so I avoided that this time.  

But since we're _only_ using `tech` from the button options, I'll change so we're not passing everything.

caseybaggz (approved):

PR #337, opened 2022-06-07
Fix/336 svelte circular progress

### LINKED ISSUE #336: [Bug?]: Svelte CircularProgress SVGs incorrect and missing box size
### Duplicates

### Latest version

### Current behavior 😯

When using getCircularProgressProps in Svelte, the SVGs are not displaying correctly and only displays in "m" size

image

### Expected behavior 🤔

Circle should be thicker, and should indicate the percentage complete.  When selecting `xs` for size, circle should be smaller.

Size is missing because we're assigning the CSS module class, rather than the class directly.
SVG issues are due to camel-cased attributes

### Steps to reproduce 🕹

Steps:

1. Create circular progress in Svelte project
2. Set size to "xs"
3. Create another with default parameters for comparison

### Context 🔦

_No response_

### Your environment 🌎

```shell
https://codesandbox.io/s/svelte-forked-5e0ok3?file=/App.svelte
```

Rykus0:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

closes #336 

Svelte output was not rendering correctly due to react-named attributes in SVG
Only one size worked in Svelte due to an incorrect class size mapping

## What is the new behavior?

Corrects attribute names to be SVG standard when outputting for Svelte
Correctly outputs the 'size' class for Svelte


packages/headless-styles/src/components/CircularProgress/shared.ts
```diff
@@ -33,17 +33,30 @@ function getDashArray(nowValue: number) {
 // public
 
 export const VIEWBOX = '0 0 100 100'
-export const baseCircleProps = {
-  cx: '50',
-  cy: '50',
-  r: '42',
-  strokeWidth: '12px',
+export function getBaseCircleProps(tech: Tech) {
+  const strokeWidthProp = tech === 'svelte' ? 'stroke-width' : 'strokeWidth'
```
caseybaggz, line 37: Hey! One of our 2 dependencies can manage this logic `kebab-case`.  You could even make 2 helpers since there is a ton of `tech` checking. Something like...

1. `isSvelteTech(tech: Tech): boolean`
2. `getCSSProperty(jsPropName: string, isSvelteTech: boolean): string`

### `getCSSProperty`

Probably something to this affect (fine-tune where needed)
```
export function getCSSProperty(jsPropName: string, isSvelte: boolean) {
  if (isSvelte) {
    return kebabCase(jsPropName)
  }

  return jsPropName
}
```

Then you can just do this
```
const strokeWidth = getCSSProperty('strokeWidth', isSvelteTech(tech))
```

All theory. Up to you on everything. I would probably for sure make the tech helper though because I feel like we are using `tech === 'svelte'` in a lot of places.

packages/headless-styles/src/components/CircularProgress/shared.ts
```diff
@@ -33,17 +33,30 @@ function getDashArray(nowValue: number) {
...
+    r: '42',
+    [strokeWidthProp]: '12px',
+  }
 }
 
-export function getStrokeProps(now: number) {
+export function getStrokeProps(now: number, tech: Tech) {
+  const dashArray = getDashArray(now)
+
+  if (tech === 'svelte') {
```
caseybaggz, line 50: Ha - like...here.

packages/headless-styles/src/components/CircularProgress/shared.ts
```diff
@@ -33,17 +34,31 @@ function getDashArray(nowValue: number) {
 // public
 
 export const VIEWBOX = '0 0 100 100'
-export const baseCircleProps = {
-  cx: '50',
-  cy: '50',
-  r: '42',
-  strokeWidth: '12px',
+export function getBaseCircleProps(tech: Tech) {
+  const strokeWidth = getAttribute('strokeWidth', getAttributeStandard(tech))
```
caseybaggz, line 38: So, here I do prefer the original naming I recommended `getCSSProperty` and `isSvelteTech` specifically because of a couple of reasons:

1. CSS properties are not attributes - and we are technically not using them ever as attributes (html tag properties)
2. getAttrStandard is really vague because I'm not sure what that means by glancing at it for the rason above. This really gets meta because technically React doesn't use html tags - because everything is JSX, attributes are called properties - even more specifically because all React components are just objects, so everything you pass to a component is accurately just a property on the component object tree. Same with the other frameworks - so this naming gets messy really fast.

packages/headless-styles/src/utils/helpers.ts
```diff
@@ -100,3 +100,13 @@ export function transformStyles(styleObject: StyleProps) {
     .trim()
     .replace(/^ {2,12}/gm, '')
 }
+
+export type AttributeStandard = 'jsx' | 'html'
+
+export function getAttributeStandard(tech: Tech) {
```
caseybaggz, line 106: Seeing this code, I would even say that it could be renamed probably `getSyntaxStandard`, then the other being `transformCasing` which ensures people who have no context of this logic know exactly what is happening.

caseybaggz (commented):
Final decision: `getSyntaxStandard` and `transformCasing`

packages/headless-styles/src/utils/helpers.ts
Rykus0: Ok.  Part of my thought around this is that it will also be used to transform ‘htmlFor’ to ‘for’ in the very near future, so it’s a little bit more than the casing.  Maybe `transformSyntax` to match `getSyntaxStandard`?  Unless that's adding too much into one method

packages/headless-styles/src/components/CircularProgress/shared.ts
Rykus0: Yeah, I figured prop/property would fit better with the React naming standards, but I went with attribute since they represent HTML attributes (not CSS in this case) at a more abstract level.  But it's also somewhat out of sync with the rest of our naming and react-first approach, so I wasn't sure about it.  Thanks for the feedback

packages/headless-styles/src/utils/helpers.ts
caseybaggz: Ah, in that case, I would definitely keep it `transformCasing` and leave the conversion of properties to HTML attrs for another function to which we can then name `jsxToHtml` or something like that. 👍

packages/headless-styles/src/utils/helpers.ts
```diff
@@ -100,3 +100,13 @@ export function transformStyles(styleObject: StyleProps) {
     .trim()
     .replace(/^ {2,12}/gm, '')
 }
+
+type Syntax = 'jsx' | 'html'
+
+export function getSyntaxStandard(tech: Tech) {
```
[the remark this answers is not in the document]
Rykus0: Definitely not married to it.  I was trying to think how to represent the abstraction and was thinking that Svelte uses html standards in its naming, whereas for React the standard is jsx. But if that's not clear, I'm happy to change it.

packages/headless-styles/tests/utils/helpers.test.ts
```diff
@@ -112,4 +114,17 @@ describe('helpers', () => {
...
     })
   })
+
+  test('getSyntaxType returns syntax based on tech', () => {
+    expect(getSyntaxType('svelte')).toEqual('html')
+    expect(getSyntaxType('')).toEqual('jsx')
+  })
+
+  test('transformCasing converts camel to kebab case when syntax is "html"', () => {
+    expect(transformCasing('ariaLabel', 'html')).toEqual('aria-label')
```
caseybaggz, line 124: Don't worry about this now, but in the future, I would like to see more verbose testing. So, the idea is to validate all the scenarios vs. *only* the "sunny paths" (what we are doing here). I know using TS helps monitor this for us, but we should never have tests that just have one `expect` which indicates we aren't properly testing all the scenarios correctly.

For example, this scenario should also test using 'jsx', and using kebab as the argument at the least.

caseybaggz (approved):

PR #358, opened 2022-06-10
Docs/icon usage improvement

### LINKED ISSUE #325: [Docs?]: update Icons Web docs to be more useful
### Latest version

### Summary 💡

The icon examples on the Usage (web) page are not as easy to consume/use as they could be. We need to do the following:

1. Show a live example for BasicExample
2. Update the basic code to *only* include icon content
3. Update basic full preview to *only* include a Icon component (i.e. remove "..." and "div" unless it's necessary to have it in the component - basically make the full preview copy/pastable for instant working in someone's app.
4. Add live example to Icon colors
5. Do all the same as #3
6. Add live example to Icon sizes
7. Change wording in "Importing by framework" section to "We export React components by default, ...". Then in the "other section - be specific to what other exports are available (i.e. svelte, svg)
8. Kill supported frameworks section
9. Kill importing SVG files - if something is marked "danger" in our docs - we should not be promoting it's use on the site.

### Motivation 🔦

_No response_

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

closes #325 

## What is the new behavior?

Improves Icon examples, removes unnecessary documentation and streamlines the rest

### DIFF
--- a/website/docs/development/packages/icons/usage.mdx
+++ b/website/docs/development/packages/icons/usage.mdx
@@ -70,17 +70,17 @@ theming while still allowing for control over the icon's color directly.
 
 <IconColor />
 
-<LiveExampleTabs
-  fullCode={<IconColorFullPreview />}
-  sandboxList={{
-    chakra: 'basic-icons-chakra-vnew2e',
-    joy: 'basic-icons-mui-q0ww53',
-    styled: 'basic-icon-css-in-js-sc-rk9wxq',
-    svelte: 'basic-icons-svelte-r0dtoj',
+```jsx
+<div
+  style={{
+    color: 'var(--ps-danger-text)',
+    backgroundColor: 'var(--ps-danger-background-weak)',
   }}
 >
-  <IconColorPreview />
-</LiveExampleTabs>
+  <WarningIcon {...getIconProps} />
+  &nbsp; Watch for demogorgons.
+</div>
+```
 
 ## Icon sizes
 
--- a/website/docs/development/packages/icons/usage.mdx
+++ b/website/docs/development/packages/icons/usage.mdx
@@ -137,8 +137,12 @@ import { MenuIcon } from '@pluralsight/icons/svelte'
 
 If needed, you can import SVGs directly. The approach will vary based on your framework and tooling.
 
-```javascript title="Importing SVG in React with create-react-app"
-import { ReactComponent as MenuArrowLeftIcon } from '@pluralsight/icons/svg/navigation/menu-arrow-left.svg'
+```jsx
+import menuArrowLeftSvg from '@pluralsight/icons/svg/navigation/menu-arrow-left.svg'
+
+function MenuIcon() {
+  return <img alt="" src={menuArrowLeftSvg} />
+}
 ```
 
 :::note
--- a/website/docs/development/packages/icons/usage.mdx
+++ b/website/docs/development/packages/icons/usage.mdx
@@ -9,6 +9,7 @@ import {
   BasicIconPreview,
   BasicIconFullPreview,
 } from '@site/src/components/Icons/BasicIcon.preview'
+import IconSizes from '@site/src/components/Icons/IconSizes'
 import {
   IconSizesPreview,
   IconSizesFullPreview,
@@ -84,6 +85,8 @@ theming while still allowing for control over the icon's color directly.
 ## Icon sizes
 
 `Icons` come in three size variants: s, m (default), l.
+
+<IconSizes />
 
 <LiveExampleTabs
   fullCode={<IconSizesFullPreview />}
--- a/website/docs/development/packages/icons/usage.mdx
+++ b/website/docs/development/packages/icons/usage.mdx
@@ -4,6 +4,7 @@ tags: [Development, Usage, Icons, Web]
 ---
 
 import LiveExampleTabs from '@site/src/components/LiveExampleTabs/LiveExampleTabs'
+import BasicIcon from '@site/src/components/Icons/BasicIcon'
 import {
   BasicIconPreview,
   BasicIconFullPreview,
@@ -36,6 +37,8 @@ npm install @pluralsight/icons@alpha
 ## Basic example
 
 Icons are provided as either SVG or component exports based on your choosing.
+
+<BasicIcon />
 
 <LiveExampleTabs
   fullCode={<BasicIconFullPreview />}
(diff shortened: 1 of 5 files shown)


website/src/components/Icons/BasicIcon.js
```diff
@@ -1,18 +1,14 @@
...
 const psIconProps = getIconProps()
-const psLargeIconProps = getIconProps({ size: 'l' })
 
 function BasicIcon() {
   return (
     <Container>
-      <MegaphoneIcon {...psSmallIconProps} />
-      <MegaphoneIcon {...psIconProps} />
-      <MegaphoneIcon {...psLargeIconProps} />
+      <MenuIcon {...psIconProps} />
```
caseybaggz, line 11: This probably looks really weird in the UI with only a single icon. I would make it 3 again, but just rando others and not the same one.

website/src/components/Icons/BasicIcon.preview.js
```diff
@@ -20,6 +12,10 @@ export function BasicIconFullPreview() {
     <CodeBlock>{`import { MenuIcon } from '@pluralsight/icons'
 import { getIconProps } from '@pluralsight/headless-styles'
 
-${preview}`}</CodeBlock>
+export default function PSMenuIcon() {
```
caseybaggz, line 15: Kill the "PS" for the component name. We don't want to assume people want it attached in case of copy/paste. It also makes the name harder to read.

website/src/components/Icons/IconColor.js
```diff
@@ -0,0 +1,29 @@
...
+          backgroundColor: 'var(--ps-danger-background-weak)',
+          display: 'flex',
+          alignItems: 'center',
+          padding: '1em',
+          borderRadius: '3px',
+          width: '100%',
+        }}
+      >
+        <WarningIcon {...psIconProps} />
+        &nbsp; This is a warning!
```
caseybaggz, line 23: Copywriting 101 - never ever use "!". We don't ever want to yell at people. 😉 Instead, try to match the new brand by being more holistic and human.

```suggestion
        &nbsp; Watch for demogorgons.
```

website/src/components/Icons/IconColor.preview.js
```diff
@@ -1,13 +1,14 @@
...
-  <WarningIcon />
-  ...
+const preview = `<div
+  style={{
+    color: 'var(--ps-danger-text)',
+    backgroundColor: 'var(--ps-danger-background-weak)',
+  }}
+>
+  <WarningIcon {...getIconProps} />
+  &nbsp; This is a warning!
```
caseybaggz, line 11:
```suggestion
  &nbsp; Thanos has all the stones.
```

website/src/components/Icons/BasicIcon.js
Rykus0: Yeah, I thought so too.  I changed it because I was basically doing the same as sizes, but showing a couple different icons is a good idea, so I'll do that.

website/src/components/Icons/BasicIcon.preview.js
Rykus0: Makes sense.  I couldn't use 'MenuIcon' since that's the name of the imported icon, so I'll go with `NavMenuIcon`

website/docs/development/packages/icons/usage.mdx
```diff
@@ -114,47 +123,26 @@ interface IconSizeGuide {
...
+If needed, you can import SVGs directly. The approach will vary based on your framework and tooling.
 
-- React
-- Svelte
-
-If you would like to request an additional format, please follow our [contribution process](../../discover-more/contributing.md#contributing).
-
-## Importing SVG files (fallback)
+```javascript title="Importing SVG in React with create-react-app"
+import { ReactComponent as MenuArrowLeftIcon } from '@pluralsight/icons/svg/navigation/menu-arrow-left.svg'
```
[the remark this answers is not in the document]
Rykus0: Ok.  I didn't know we could do this without specific tooling.  I was also trying to avoid using the icons in an `img` since we lose some features that way. I want to refresh my memory on it in case we need to add any caveats for this approach.  But I'll push the updates in the meantime, and we can address it later if needed.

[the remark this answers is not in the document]
Rykus0: Hm.  Right now this isn't working I believe because we strip the xmlns attribute off, which is needed in some cases.  We might have to leave it in for the exported svgs, and remove it only when we have to.

website/docs/development/packages/icons/usage.mdx
```diff
@@ -114,47 +123,30 @@ interface IconSizeGuide {
...
-- React
-- Svelte
+```jsx
+import menuArrowLeftSvg from '@pluralsight/icons/svg/navigation/menu-arrow-left.svg'
 
-If you would like to request an additional format, please follow our [contribution process](../../discover-more/contributing.md#contributing).
-
-## Importing SVG files (fallback)
+function MenuIcon() {
+  return <img alt="" src={menuArrowLeftSvg} />
```
caseybaggz, line 144:
```suggestion
  return <img alt="menu arrow left" src={menuArrowLeftSvg} />
```

website/src/components/Icons/IconColor.preview.js
```diff
@@ -17,9 +18,25 @@ export function IconColorPreview() {
...
     <CodeBlock>{`import { WarningIcon } from '@pluralsight/icons'
+import { getIconProps } from '@pluralsight/headless-styles'
 
-const DangerBlock = (
-${preview.replace(/^(.)/gm, '  $1')}
-)`}</CodeBlock>
+function Warning(props) {
+  return (
+    <div
+      style={{
```
caseybaggz, line 26: In our previews, we never want to show all the "extra junk" we are doing behind the scenes - so this means basically the `style` prop. 

If an API requires these updates in order to do what you are promoting...then the API should be handling it on behalf of the user - not the preview.

However, on this topic - our Webkit doesn't require changing the colors of icons - which means we probably shouldn't host a color example section like you are doing here. Instead of doing a "live example" component with previews - just do a code snippet (like we do in FormControl).

Rykus0: I don't think I completely understand this, so let me know if I've missed the mark on my updates.

caseybaggz (approved):

website/src/components/Icons/IconColor.preview.js
caseybaggz: perfect!

Issue #363, opened 2022-06-13
[Feature?]: Configurable SVG Strings


Rykus0:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

This method should accept a parameter or parameters to modify the SVG code in specific ways and return it as a string.  To begin with, this will just be color.  We are assuming an embedded usage by default, so we format as a data URL and remove unnecessary attributes.

```javascript
getIconString(name, color)
```

`name` = name of the icon as per other exports (e.g., "check.svg" == "CheckIcon")
`color` = a CSS color value

### Examples 🌈

```
import { getIconString } from '@pluralsight/icons'

getIconString('CheckIcon', '#fff')
```

returns

`'data:image/svg+xml,<svg viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg" fill="#fff"><path d="m9.59 14.58-2.818-2.818a.5.5 0 0 0-.706 0l-.71.704a.5.5 0 0 0-.003.708h.002l3.881 3.882a.5.5 0 0 0 .707 0l9.293-9.292a.5.5 0 0 0 0-.708l-.703-.702a.5.5 0 0 0-.707 0L9.59 14.58Z"/></svg>'`

### Motivation 🔦

In some cases, you need to embed an SVG as a string (img src attribute, CSS background-image, etc.) In these situations, some of the assumed benefits of the inline SVG so not work the same, such as inheriting color, so we will need to modify the SVG code that is returned.

PR #365, opened 2022-06-13
Refactor/icon iterator

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

Contributes to #363 

## What is the new behavior?

Refactor icon build methods to use a common method to iterate over SVGs and use for Svelte output generation

## Other information

This is to help make our build process simpler and more scalable.

### DIFF
--- a/packages/icons/utils/svgToReact.mjs
+++ b/packages/icons/utils/svgToReact.mjs
@@ -2,29 +2,18 @@ import path from 'path'
 import fs from 'fs'
 import { transform } from '@svgr/core'
 import svgrOptions from '../svgr.config.cjs'
+import iterateSvgs from './iterateSvgs.mjs'
 
 const srcPath = path.join('build', 'svg')
 const buildRoot = path.join('build', 'generated')
 const buildPath = path.join(buildRoot, 'react')
 const indexFile = path.resolve(buildRoot, 'index.ts')
 
-function toPascalCase(name) {
-  return name
-    .replace(/-(\w)/g, (x) => x.toUpperCase())
-    .replace(/-/g, '')
-    .replace(/^([a-z])/, (x) => x.toUpperCase())
+function getOutputDir(pathName) {
+  return pathName.replace(srcPath, buildPath)
 }
 
-function formatAndWriteTsx(reactIconContent, varName, outputPath) {
-  const outputFile = `${varName}.tsx`
-
-  fs.writeFileSync(path.resolve(outputPath, outputFile), reactIconContent)
-
-  fs.appendFileSync(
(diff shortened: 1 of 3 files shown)


packages/icons/utils/svgToReact.mjs
```diff
@@ -38,45 +27,49 @@ function formatAndWriteTsx(reactIconContent, varName, outputPath) {
...
+
+  const dirIndexFile = path.resolve(outputPath, pathName, 'index.ts')
+  if (fs.existsSync(dirIndexFile)) {
+    fs.rmSync(dirIndexFile)
+  }
 }
-svgToReact(srcPath)
+
+console.log('Generating React icons...')
+removeIndexFile()
```
caseybaggz, line 73: Great function names on this page! So easy to read and understand at a glance!

packages/icons/utils/svgToSvelte.mjs
```diff
@@ -1,67 +1,56 @@
 import path from 'path'
 import fs from 'fs'
+import iterateSvgs from './iterateSvgs.mjs'
 
 const srcPath = path.join('build', 'svg')
```
caseybaggz, line 5: I would probably code split `.join` at this point since you are using it so much.

```
import { join } from 'path' 
```

packages/icons/utils/svgToSvelte.mjs
```diff
@@ -1,67 +1,56 @@
...
+    fs.rmSync(indexFile)
+  }
 }
 
-if (fs.existsSync(indexFile)) {
-  fs.rmSync(indexFile)
+function createOutputDir(pathName) {
+  const outputPath = getOutputDir(pathName)
+
+  if (!fs.existsSync(path.resolve(outputPath))) {
```
caseybaggz, line 48: Normally we have an eslint rule (I keep forgetting to add it) which prevents the use of negated conditions because they are harder to read (i.e. make your brain work unnecessarily harder to interpret) and increase the likelihood of bugs.

Having said that, if there is a way we can reverse the logic to be able to remove it - we should.

Rykus0: I went with a try/catch, which I've often heard as the preferred method for these synchronous node methods.  I logged a message regarding the error, though considered either setting it as a warning (since it might not be a problem) or ignoring it (since it likely already exists), but didn't since it could be other issues.  Might be better to log the error itself... I can update if needed.

caseybaggz (approved):

Issue #379, opened 2022-06-16
[Feature?]: add getTextareaProps CSS & JS API


caseybaggz:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

- [x] CSS
- [x] JS
- [x] SC SB
- [x] Svelte SB
- [x] Chakra SB
- [x] MUI SB

### Examples 🌈

![Screen Shot 2022-06-16 at 8 49 33 AM](https://user-images.githubusercontent.com/4819738/174084634-f781ca5a-d5a5-46bc-9569-1eea2973f64c.png)

### Motivation 🔦

https://www.figma.com/file/ZmH4XsZS5WnKeo28ylM5x1/PS-Design---Web-UI-Kit-%5BALPHA%5D?node-id=1215%3A55072

PR #382, opened 2022-06-16
Feat/379 textarea CSS API

caseybaggz:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

contributes #379 

## What is the new behavior?
Adds Textarea CSS api and sandbox example


packages/headless-styles/src/components/Textarea/textareaCSS.module.css
```diff
@@ -0,0 +1,86 @@
+.textareaBase {
+  appearance: none;
+  background: var(--ps-neutral-surface-weak);
+  border-color: var(--ps-neutral-border);
+  border-image: initial;
+  border-radius: 6px;
+  border-style: solid;
+  border-width: 1px;
+  color: white;
```
Rykus0, line 9: Text isn't readable in light mode.

packages/headless-styles/src/components/Textarea/shared.ts
```diff
@@ -0,0 +1,50 @@
...
+    resize: options?.resize ?? defaultTextareaOptions.resize,
+    tech: options?.tech ?? defaultTextareaOptions.tech,
+    value: options?.value ?? defaultTextareaOptions.value,
+  }
+}
+
+export function createTextareaOptions(options: TextareaOptions) {
+  const { describedBy } = options
+  const a11yProps = createA11yProps(options)
+  const describedByProps = describedBy && {
```
Rykus0, line 38: I think we should combine this with `a11yProps` if we can

packages/headless-styles/tests/textarea/textareaCSS.test.ts
```diff
@@ -0,0 +1,65 @@
...
+  test('should accept a resize type', () => {
+    expect(getTextareaProps({ ...options, resize: 'initial' })).toEqual(result)
+    expect(getTextareaProps({ ...options, resize: 'horizontal' })).toEqual({
+      ...result,
+      className: `${baseClass} horizontalTextarea`,
+    })
+    expect(getTextareaProps({ ...options, resize: 'vertical' })).toEqual({
+      ...result,
+      className: `${baseClass} verticalTextarea`,
+    })
```
Rykus0, line 45: Missing the 'none' case

packages/headless-styles/src/components/Textarea/textareaCSS.module.css
caseybaggz: Alas...finally having no white variable is an issue. 😭 Guess I'll just use whatever we are using on Input...not sure why it's different for TA anyways? 🤷‍♀️

packages/headless-styles/tests/textarea/textareaCSS.test.ts
caseybaggz: Good eye!

packages/headless-styles/src/components/Textarea/shared.ts
caseybaggz: @Rykus0 there is a reason we don't do that in `createA11yProps` which are that not all inputs/components using that function require the `describedBy` attr. So, adding that logic into the function just for a couple of use cases is by nature a bad functional programming and clean code practice. It's better to keep it separate so the function stays pure which will reduce the risk of bugs happening over the lifetime of the code.

Rykus0 (approved):

Issue #408, opened 2022-07-06
[Feature?]: Add Tag CSS & JS API


Rykus0:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

- [x] CSS
- [x] JS
- [x] Svelte
- [x] SC
- [x] Chakra
- [x] MUI

PR #421, opened 2022-07-11
Feat/get tag props

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

contributes to #408 

## What is the new behavior?
Adds tag CSS API and tests


packages/headless-styles/sandbox/src/App.css
```diff
@@ -15,8 +15,12 @@
 }
 
 .App-header {
+  background: #fff;
```
caseybaggz, line 18:
```suggestion
  background-color: var(--ps-neutral-surface-medium);
```

packages/headless-styles/sandbox/src/App.css
```diff
@@ -15,8 +15,12 @@
...
   padding-bottom: 1.5rem;
   padding-top: 1.5rem;
+  position: sticky;
+  top: 0;
+  z-index: 50;
```
caseybaggz, line 23:
```suggestion
  z-index: 1100;
```

packages/headless-styles/src/components/Tag/generated/tagCSS.module.ts
```diff
@@ -0,0 +1,79 @@
...
+    '&:active': {
+      backgroundColor: 'hsl(240deg 31% 25% / 100%)',
+    },
+  },
+  sTag: {
+    composes: 'baseTag',
+    fontSize: '0.75rem',
+    height: '1.25rem',
+  },
+  '': {
```
caseybaggz, line 69: Hmmm...this looks like a bug to me. Need to figure out what the CSS file needs to change in order to prevent the empty selector.

packages/headless-styles/src/components/Tag/shared.ts
```diff
@@ -0,0 +1,16 @@
+import type { Tech } from '../types'
+import type { TagOptions, Kind, Size } from './types'
+
+export const defaultTagOptions = {
+  kind: 'weak' as Kind,
+  size: 'm' as Size,
+  tech: '' as Tech,
+}
+
+export function getDefaultOptions(options?: TagOptions) {
```
caseybaggz, line 10: Need to rename this function to use the pattern `getDefault<Name>Options` since this is a global function.

packages/headless-styles/src/components/Tag/tagCSS.module.css
```diff
@@ -0,0 +1,85 @@
...
+  position: relative;
+  text-align: center;
+  text-decoration: none;
+  text-transform: none;
+  transition: background-color 250ms ease-in-out, color 250ms ease-in-out;
+  vertical-align: middle;
+  white-space: nowrap;
+}
+
+[href].baseTag {
```
caseybaggz, line 27: We want to stick with class name selectors and not strict attrs like this one. Theorhetically, you should only need `.baseTag` or something different to make it unique enough.

packages/headless-styles/src/components/Tag/tagCSS.module.css
```diff
@@ -0,0 +1,85 @@
...
+}
+
+.weakTag:active {
+  background-color: var(--ps-neutral-background-active);
+}
+
+/* focus styles */
+
+.baseTag:focus {
+  outline: hsl(0deg 0% 100% / 80%) solid 2px;
```
caseybaggz, line 72: All focus styles should match the other components and not use static colors which will also allow you to kill the global style at the bottom of the file. Always reference the other components since the library is essentially complete and not "go rougue" - even if the design file is currently different (which is expected because it in progress).

packages/headless-styles/src/components/Tag/tagCSS.ts
```diff
@@ -0,0 +1,20 @@
+import { createClassProp } from '../../utils/helpers'
+import { getDefaultOptions } from './shared'
+import type { TagOptions } from './types'
+import styles from './tagCSS.module.css'
+
+// Public
```
caseybaggz, line 6:
```suggestion
```

packages/headless-styles/src/components/Tag/shared.ts
Rykus0: oops, thanks!

packages/headless-styles/src/components/Tag/generated/tagCSS.module.ts
Rykus0: I bet it's because of `:global` - I noticed the theming wasn't working because CSS modules was changing the `light` class name the way it does.  This looks like it introduces some other issues, though, so I'll have to look more into it.

packages/headless-styles/src/components/Tag/tagCSS.module.css
Rykus0: I was trying to target only when it's an active link with an href... I could put this in with base if that's always the case.

packages/headless-styles/src/components/Tag/tagCSS.module.css
Rykus0: I had that initially, but it didn't match what we had in figma, and what's existing is hard to see.  I was going to ask if we wanted to either add these colors to the tokens, or use existing colors, but forgot to come back to it.  I'll make sure to get clarification on the expectation next time I have any degree of doubt.

Rykus0: Does css modules have something like SASS mixins that could help us keep things like this consistent more easily?

packages/headless-styles/src/components/Tag/generated/tagCSS.module.ts
Rykus0: No longer an issue with `:global(.light)` removed

caseybaggz (approved):
I don't understand how there is not an aria selector for Tags at this point? They've been around for like a decade now? 🤷‍♀️

PR #423, opened 2022-07-12
Feat/get js tag props

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

contributes to #408 

## What is the new behavior?

adds getJSTagProps API


Rykus0:
[SC Sandbox](https://codesandbox.io/s/basic-tag-css-in-js-sc-ykiwwt?file=/src/App.tsx) 
[MUI Sandbox](https://codesandbox.io/s/basic-tag-mui-xdhfm8?file=/src/App.tsx) - will provide override in following PR along with Chakra API

packages/headless-styles/sandbox/src/components/Tag.jsx
```diff
@@ -1,6 +1,18 @@
...
+  useEffect(() => {
+    if (props.logJS) {
+      console.log(
+        getJSTagProps({
+          kind: 'strong',
+          size: 's',
+        })
+      )
+    }
+  }, [props.logJS])
```
caseybaggz, line 14: Just a fun fact via React knowledge - what you assume is correct in the `props.logJS` being used in the Array of props to be checked is in fact **incorrect** because React *only* checks the Array on a shallow level. Thus, the effect will run every single time the `props` Object changes - not the `logJS` property.

This is why all the React docs and the code I write use the `logJS` property specifically. A common issue for non-contributing React devs. Might change in the future - but I doubt it.

packages/headless-styles/src/components/Tag/tagJS.ts
```diff
@@ -0,0 +1,22 @@
+import { getDefaultTagOptions } from './shared'
+import { transformStyles } from '../../utils/helpers'
+import type { TagOptions } from './types'
+import styles from './generated/tagCSS.module'
+
+// Public
```
caseybaggz, line 6: Please stop adding these comments if there is no reason for it. 😉 
```suggestion
```

Rykus0: Didn't really add it, I used other code as a starting point and forgot to remove it.

packages/headless-styles/src/components/Tag/tagJS.ts
```diff
@@ -0,0 +1,21 @@
+import { getDefaultTagOptions } from './shared'
+import { transformStyles } from '../../utils/helpers'
+import type { TagOptions } from './types'
+import styles from './generated/tagCSS.module'
+
+
+export function getJSTagProps(options?: TagOptions) {
+  const { kind, size } = getDefaultTagOptions(options)
+  const sizeKey = `${size}Tag` as keyof typeof styles
```
caseybaggz, line 9: I added a new design that helped me last API if you want to pick up on your next one (not required, but highly recommend) of creating `shared/createSelectorClasses` function that returns an Object of these values since you use them in both the CSS & JS API.

Rykus0: Makes sense.  I'll add it in the next PR!

PR #462, opened 2022-07-26
Docs/avatar

### LINKED ISSUE #447: [Docs?]: Avatar headless API documentation
### Latest version

### Summary 💡

- [ ] CSS
- [ ] JS

### Motivation 🔦

_No response_

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

closes #447 

## What is the new behavior?

Adds Avatar API documentation with examples

## Other information

Fixes an incorrect name in the basic Tag documentation examples


website/docs/development/headless-styles/Avatar.mdx
```diff
@@ -0,0 +1,180 @@
...
+
+:::tip
+
+We use the standard clothing size shorthand values for all `size` related options in any library we provide (i.e. xs-xxl).
+
+:::
+
+### Avatar Content
+
+An Avatar can display a profile image, initials, or a fallback icon.
```
caseybaggz, line 93: I would remove the example here and instead focus on describing *when* to use each content. Should be visually appealing too, so maybe an intro section, then use h4's for each content style and describe when it is appropriate to use. Then wrap it up with a warning admonition in the "placeholder" and "text" stating that "X" style should not be the default option displayed.

I would probably title them in a specific order too. Something like:

#### Image (preferred)
blah blah

#### Inintials (fallback)
blah blah

:::warning
should only be used as a fallack to image, blah blah
:::

#### Icon (fallback - last resort)
blah blah

:::danger
should only be used when there is 1. No avatar image 2. no initials to be used. This should be a last resort blah blah
:::

website/src/components/Avatar/AvatarContent.js
```diff
@@ -0,0 +1,61 @@
...
+  return <FallbackAvatar {...props} />
+}
+
+function AvatarContent() {
+  return (
+    <Container>
+      <Avatar
+        name="Weird Al Yankovic"
+        src={weirdAl}
+        ariaLabel="Weird Al's Profile"
```
caseybaggz, line 49: Let's not use Weird Al in our public examples (sandboxes are fine). Instead, try and find an avatar of a non-famous person because famous people always get in trouble at some point in time and we don't ever want to be on the side of "support" if/when that happens. 😉

website/docs/development/headless-styles/Avatar.mdx
Rykus0: Awesome, thanks!  I was going to do something along these lines but didn't want to get too verbose.  But I also forgot about the direction for when to use each! which we definitely need!

caseybaggz (approved):

PR #464, opened 2022-07-27
Feat/get text link props

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

contributes to #418 

## What is the new behavior?

Adds CSS API for TextLink


Rykus0:
[TS React](https://codesandbox.io/s/basic-text-link-react-typescript-ytktg8?file=/src/App.tsx)
[Svelte](https://codesandbox.io/s/basic-text-link-svelte-r816y1?file=/App.svelte)

packages/headless-styles/sandbox/src/components/TextLink.jsx
```diff
@@ -0,0 +1,33 @@
...
+  const { link, iconOptions } = getTextLinkProps()
+
+  return (
+    <div id="text-link">
+      <h3>TextLink</h3>
+      <div className="App-container">
+        <a href="#text-link" {...link}>
+          text link
+        </a>
+        <a href="https://www.google.com" {...link}>
```
caseybaggz, line 18: I would actually update this API to include the `href` as a property. Then, you can internally check if the string contains the characters `http` and return all the required a11y link props (target, rel). It should be easier on the jsx as well.

This:
```
<a {...linkProps} />
```

Is much cleaner than this:
```
<a href="https://www.google.com" {...link}>
```

Just a smaller step to allow HS to own a little more since inline links can be either relative or external.

packages/headless-styles/src/components/Icon/iconCSS.module.css
```diff
@@ -19,3 +19,7 @@
   height: 1.5rem;
   width: 1.5rem;
 }
+
+.customSize {
```
caseybaggz, line 23: This isn't going to be a scalable class name due to the nature of how Svelte works. It's fine for every thing, except Svelte because they don't use modules. 😭 

I would recommend changing all the classes to use `.<ElementName><className>`. Totally unnecessary for CSS modules, but, because we want to support svelte - we needs to do it.

packages/headless-styles/src/components/TextLink/textLinkCSS.module.css
```diff
@@ -0,0 +1,37 @@
...
+    sans-serif;
+  font-variation-settings: 'wght' 400;
+  font-weight: 400;
+  gap: 0.25em;
+  outline: none;
+  text-decoration: underline;
+  transition: background-color 250ms ease-in-out, color 250ms ease-in-out;
+}
+
+.baseTextLink:visited {
```
caseybaggz, line 17: See above comment. This language is close, but not the pattern we are using. Should instead be `.textLinkBase`

Rykus0: Ok, I'd been using ${class}${component-specific-suffix} to make it easier for appending to anything without needing to fiddle with the case.  I can switch still it, but I'd have to change the other ones I did, too (at some point).

packages/headless-styles/sandbox/src/components/TextLink.jsx
Rykus0: Ok.  I'd thought about that at first, but thought it was outside of our scope, though I can also see how it can be within those bounds.

packages/headless-styles/src/components/Icon/iconCSS.module.css
Rykus0: Ah, yeah, my mistake.  It should be `.customIconSize`

packages/headless-styles/sandbox/src/components/TextLink.jsx
```diff
@@ -0,0 +1,48 @@
...
+  return (
+    <a {...link}>
+      {IconBefore && <IconBefore {...getIconProps(iconOptions)} />}
+      {props.children}
+      {IconAfter && <IconAfter {...getIconProps(iconOptions)} />}
+      {link.rel?.includes('external') && (
+        <ExternalLinkIcon
+          {...getIconProps({
+            ...iconOptions,
+            ariaHidden: false,
```
caseybaggz, line 23: This should be included in the API. We should only force users to extend if they want to actually extend base features. Since we know now if a link is external, there's no reason for us to not update the iconOptions internally so the user doesn't have to do this.

packages/headless-styles/src/components/TextLink/shared.ts
```diff
@@ -0,0 +1,36 @@
+import type { Tech } from '../types'
+import type { TextLinkOptions } from './types'
+
+function isExternalHref(href: string) {
+  return /^https?:\/\//.test(href)
+}
+
+function getRel(href: string) {
+  return isExternalHref(href) ? 'noopener external' : ''
```
caseybaggz, line 9: All external links should use `target="_blank"` as well. This is a standard pattern now.

packages/headless-styles/src/components/TextLink/shared.ts
```diff
@@ -0,0 +1,36 @@
...
+    href: options?.href ?? defaultTextLinkOptions.href,
+    tech: options?.tech ?? defaultTextLinkOptions.tech,
+  }
+}
+
+export function createTextLinkProps(href: string, tech: Tech) {
+  return {
+    link: {
+      href: href,
+      rel: getRel(href),
```
caseybaggz, line 28: We shouldn't return this property if it's not being used. Suprised to see this pattern used by you since in previous API's you have done that. 🤔

packages/headless-styles/src/components/TextLink/types.ts
```diff
@@ -0,0 +1,6 @@
+import type { Tech } from '../types'
+
+export interface TextLinkOptions {
+  href?: string
```
caseybaggz, line 4:
```suggestion
  href: string
```

packages/headless-styles/src/components/TextLink/shared.ts
Rykus0: That contradicts what I know in terms of UX and A11y, which is that it should only be used if following the link would result in a loss of data. So in that case it would be on the application side.

packages/headless-styles/src/components/TextLink/shared.ts
Rykus0: Sorry, I got confused.  I was going to by returning `undefined` from `getRel` but remembered you had said not to do that.  I can do it that way, or a different way.  I would have, but I doubted myself, so I'll fix it up.

packages/headless-styles/sandbox/src/components/TextLink.jsx
Rykus0: I didn't include it by default because we won't always know if the text already describes that the link is external, so I left it out.  We can include it by default and have them override it if needed; maybe that's a better default?  At least then, the worst case is that the label is redundant as opposed to missing information.

caseybaggz: Exactly, the main point here is that for any of our native API's - it shouldn't require extra work *unless* you want to add features we don't offer or replace ours.

I definitely see this as a "why are you making me write this?" scenario.

packages/headless-styles/src/components/TextLink/shared.ts
caseybaggz: No, no - you are close but it's not fully accurate regarding UX.

**All external links should point to a new tab** - because you are theoretically choosing to opt out of one experience and into another (i.e. most external links are completely different sites so the data thing doesn't apply). This is also why we show the "external link" icon to make it known to the user that they are about to "leave our site". This also allows the user to have a safe place to return to that has not been accidentally affected by a normal "page refresh" event which is what happens when altering the history.

**All non-external links should not point to a new tab** - 100% for the reason you mentioned above.

packages/headless-styles/src/components/TextLink/shared.ts
caseybaggz: You want to use the same pattern we are already using in other helpers which only returns the property if it exists.
https://github.com/pluralsight/tva/blob/main/packages/headless-styles/src/components/Input/shared.ts#L40

packages/headless-styles/src/components/TextLink/shared.ts
Rykus0: I disagree.  What you're describing is an outdated pattern.  If the icon is for links that open in a new window (not necessarily external), then I agree.  But we're changing the expectation and the default functionality, which has negative accessibility and UX implications.  

It's fine to force some external links to open in a new window, but not all external links should by default.  

If that's the way we're doing it, though, I'll do it that way.

packages/headless-styles/sandbox/src/components/TextLink.jsx
Rykus0: I just had another thought, which is that there may be an additional icon in the link that is not the one that indicates the link opens in a new window.  So we'd either need to separate the returned props, default to the 'hidden' icon, or only allow the 'external link' icon.  Do you have a preference?  I'm leaning toward the separate props for the external icon.

caseybaggz: @Rykus0 honestly, people shouldn't be using external links without the external link icon - that's a major UX fallacy. This is also why I wanted to create the icon map so we can just tell them which Icon to use for the external link. So, I would prefer we assume the external link requires the external icon only and make that purposefully strict to ensure we are doing the web justice.

packages/headless-styles/src/components/TextLink/shared.ts
caseybaggz: A simple google search even validates this theory:

>"Keeping users in the same tab helps them better understand the navigation flow of your site. **External links, however, should open in new tabs.** These links take users to a different website. Many designers make the mistake of opening external links in the same tab."

Let's please follow that pattern.

packages/headless-styles/src/components/TextLink/shared.ts
```diff
@@ -0,0 +1,47 @@
...
+  const relProps = rel && { rel: rel }
+  const targetProps = target && { target: target }
+
+  return {
+    link: {
+      href: href,
+      ...relProps,
+      ...targetProps,
+    },
+    iconOptions: {
```
caseybaggz, line 40: Shouldn't this be optional too? I feel like it should since we only want to promote using an icon when it's external? I honestly would create a function called `getLinkProps` or something like that and `getIconOptions` that would be responsible for calling the other functions so the createTextlink is cleaner and has less logic in it.

Something like this:
```
function createTextLinkProps(href: string, tech: Tech) {
  const linkProps = getLinkProps(href)
  const iconOptionProps = isExternalHref(href) && getIconOptions()

  return {
   link: { href, ...linkProps},
   iconOptions: { ...iconOptionProps }
  }
}
```

Because you have all the other functions, no need to make createTextLinkProps own anything else outside of just being pure-ish.

Rykus0: Yeah, I had asked that, but I think it got lost in the shuffle.  I wasn't sure anymore if it was better to always have the same output shape, or to leave out the stuff that wasn't used.

Rykus0: I even had a method like this, but removed it 😆

Issue #486, opened 2022-08-11
[Feature?]: Tokens Update


caseybaggz:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

- [x] Update tokens with new category, names, values
- [x] Update all normalize references
- [x] Update Website
- [x] Update all HS references
- [x] Update tokens page

caseybaggz:
Complete. Closing.

PR #509, opened 2022-08-22
Docs/486 new token values

caseybaggz:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

contributes #486 

## What is the new behavior?
* Adds new tokens
* Updates docs to use new tokens
* Updates docs styles/design
* Updates HS package to use new token/APIs

## Other information
Will temporarily create visual bugs in HS examples.


packages/design-tokens/styles/base.css
```diff
@@ -122,6 +122,12 @@ mark::after {
   width: 1px;
 }
 
+label {
+  font-size: 0.875rem;
+  font-variation-settings: "'wght' 700";
```
Rykus0, line 127: This shouldn't be quoted, right?

```suggestion
  font-variation-settings: 'wght' 700;
```

packages/headless-styles/src/components/Avatar/shared.ts
```diff
@@ -18,32 +19,71 @@ export const iconSizeMap: Record<Size, string> = {
...
       tech: options.tech,
     },
+    wrapper: {},
+    label: {
+      'aria-label': label,
+      value: initials,
+    },
+    image: {
+      alt: label,
+      src: options.src,
```
Rykus0, line 86: I had originally hidden the image and put the aria-label on the outer element to make sure the accessible label was on the top-most, potentially focusable, element.  The image was also hidden so the alt could be the initials, that way if the image doesn't load, it's an automatic fallback.  I felt that approach was overall more flexible, leveraging fallbacks and giving more choices in the outer element used.

packages/headless-styles/src/components/Avatar/shared.ts
```diff
@@ -18,32 +19,71 @@ export const iconSizeMap: Record<Size, string> = {
...
+    firstLetter: formatLetter(firstWord),
+    lastLetter: formatLetter(labelWords[1]),
+  }
+}
+
+function formatLetter(word: string) {
+  return word?.slice(0, 1) ?? ''
+}
+
+function displayInitials(initialsObj: InitialsProps) {
```
Rykus0, line 59: I'd thought about adding something like this, but thought it was out of our bounds.

Rykus0 (approved):

PR #522, opened 2022-08-24
Fix/tag styles

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

contributes to #486 

## What is the new behavior?

Updates Tag styles to match Figma


packages/headless-styles/src/components/Tag/tagCSS.module.css
```diff
@@ -2,23 +2,23 @@
   align-items: center;
   appearance: none;
   background-color: transparent;
-  border: 2px solid var(--ps-action-navigation);
-  border-radius: 12px;
+  border: 1px solid var(--ps-border-strong);
+  border-radius: 1em;
```
caseybaggz, line 6: We've been using `px` everywhere which I think may be a little safer since BR is a little weird. Mind if we switch back to px so it's consistent?

```suggestion
  border-radius: 12px;
```

caseybaggz (approved):

packages/headless-styles/src/components/Tag/tagCSS.module.css
Rykus0: The value is relative to the font size, which is why I used `em`, but for consistency I can change it.  It means we have to add a different value for the small size, which is another reason I used the single em value.

caseybaggz (approved):

packages/headless-styles/src/components/Tag/tagCSS.module.css
caseybaggz: Ahhhh, I see - yeah, since we need a small class anyways, for the sake of keep the code simple to understand I would keep it pixels.  BR is already messy and adding `em` to it makes it even worse (especially since we are using `rems` so I can see a junior dev thinking it's a mistake and changing it.

Your rationale makes total sense and if we were a product team, I would abosolutely leave it. Being an open-source - makes it a little trickier.

Issue #502, opened 2022-08-18
[Feature?]: Tabs API


caseybaggz:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

- [x] CSS API
- [x] JS API
- [x] unit tests
- [ ] SC Sandbox
- [ ] Svelte Sandbox
- [ ] Chakra Sandbox
- [ ] MUI Sandbox

caseybaggz:
Done. Closing.

PR #621, opened 2022-09-22
Feat/502 tab api

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

contributes to #502 

## What is the new behavior?

Adds CSS and JS Tab APIs


packages/headless-styles/src/components/Tab/tabCSS.module.css
```diff
@@ -0,0 +1,101 @@
+.tabWrapper {
```
caseybaggz, line 1: Let's kill these two, but keep the objects in case we ever want to add something in the future.

Rykus0: I do have some styles going into the `.tabWrapper` with the overflow update, but nothing yet for `.tabPanelWrapper`

packages/headless-styles/src/components/Tab/tabCSS.ts
```diff
@@ -0,0 +1,52 @@
...
+      }),
+    },
+    tab: {
+      ...props.tab,
+      ...createClassProp(defaultOptions.tech, {
+        defaultClass: `${TAB} ${styles[sizeClass]}`,
+        svelteClass: `tabBase ${sizeClass}`,
+      }),
+    },
+    panelWrapper: {
```
caseybaggz, line 41: Because there is nothing here, you can kill the object so it's less lines in here. Will still automatically ship since it's included in the `...props` spread.

packages/headless-styles/src/components/Tab/tabJS.ts
```diff
@@ -0,0 +1,65 @@
...
+    },
+    tabList: {
+      ...baseProps.tabList,
+      ...createJSProps(transformStyles(jsStyles.tabList), jsStyles.tabList),
+    },
+    tab: {
+      ...baseProps.tab,
+      ...createJSProps(transformStyles(jsStyles.tab), jsStyles.tab),
+    },
+    panelWrapper: {
```
caseybaggz, line 57: Same here.

packages/headless-styles/src/components/Tab/tabCSS.ts
```diff
@@ -0,0 +1,49 @@
...
+import { getDefaultTabOptions, createTabProps } from './shared'
+import type { TabOptions } from './types'
+import styles from './tabCSS.module.css'
+
+const TAB = 'ps-tab'
+
+export function unstable_getTabProps(options?: TabOptions) {
+  const defaultOptions = getDefaultTabOptions(options)
+  const props = createTabProps()
+  const sizeClass = `${defaultOptions.size}Tab`
```
caseybaggz, line 12: Gah, totally missed this. We need to create a `getTabClasses` function to create a SSoT and stay consistent to the other APIs

Rykus0: Ah yes, thanks

caseybaggz (approved):

PR #622, opened 2022-09-26
Feat/502 chakra tabs

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

contributes to #502 

## What is the new behavior?

Adds Chakra theme for Tabs
Minor adjustments to styles:
- Apply background color to panels as well as tabs for consistency
- Use flex gap for tab spacing rather than sibling selectors for greater compatibility in applying styles across frameworks and techniques

### DIFF
--- a/packages/headless-styles/src/components/Tab/tabCSS.module.css
+++ b/packages/headless-styles/src/components/Tab/tabCSS.module.css
@@ -14,6 +14,8 @@
     transparent 100%
   );
   display: flex;
+  gap: 2rem;
+  justify-content: space-between;
   overflow-y: auto;
   padding-block-end: 1rem;
 }
@@ -34,10 +36,6 @@
   white-space: nowrap;
 }
 
-.tabBase + .tabBase {
-  margin-inline-start: 2rem;
-}
-
 .mTab {
   composes: tabBase;
 }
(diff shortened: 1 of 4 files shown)


Rykus0:
Sandbox: https://codesandbox.io/s/basic-tabs-chakra-x8lzym?file=/src/App.tsx

packages/headless-styles/src/components/Tab/tabCSS.module.css
```diff
@@ -85,6 +80,7 @@
 }
 
 .tabPanel {
+  background-color: var(--ps-surface-weak);
```
caseybaggz, line 83: This seems off. What was the reason for adding it in?

Rykus0: The tabs have that background but the panels didn't.  I noticed it in the Chakra sandbox because the app background was different, so I added this to make sure the tabs and panels matched.

caseybaggz: OK, we need to kill the tabs BGC too. We shouldn't be shipping that which explains why the Chakra sandbox looks weird.

caseybaggz: Tabs/panels should always be transparent.

Rykus0: Ok, no problem

caseybaggz (approved):

PR #637, opened 2022-09-28
Feat/tabs to next

caseybaggz:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

tabs flags in experimental channel

## What is the new behavior?
Moves to next channel.

## Other information
github-readme (1)

### DIFF
--- a/packages/shared/src/utils/helpers.ts
+++ b/packages/shared/src/utils/helpers.ts
@@ -17,3 +17,9 @@ export function createFlag(flag: Flag) {
 
   return checkFlagAgainstChannel(flag)
 }
+
+export function warning(condition: boolean, message: string) {
+  if (!condition) {
+    console.error(message)
+  }
+}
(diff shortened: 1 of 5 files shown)


packages/shared/src/utils/helpers.ts
```diff
@@ -17,3 +17,9 @@ export function createFlag(flag: Flag) {
 
   return checkFlagAgainstChannel(flag)
 }
+
+export function warning(condition: boolean, message: string) {
+  if (!condition) {
+    console.error(message)
```
Rykus0, line 23: Should we use `console.warn` instead?

caseybaggz: warn & error are just color differences in the log. We want to show `red` so we use error. Same technique as react. Function is called warning though, because we want to warn the user against stuff.

But...console wise...we want red!

Rykus0: Thanks!

Rykus0 (approved):

PR #658, opened 2022-10-04
docs: initial commit of button page

bradryanbice:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes


website/docs/design/button.md
```diff
@@ -0,0 +1,135 @@
...
+  - The default button size conforms to standard tap target sizes without a larger interactive area needed surrounding it.
+- S
+  - Use in condensed areas and smaller widgets when aligned with adjacent medium-sized form controls.
+
+### Use cases
+
+- With icons
+  - Use leading icons for presentational and ornamental glyphs only, to emphasize meaning
+  - Directional or action-oriented icons such as arrows should always be placed on the trailing side of the button
+  - Avoid using icons on both the leading and trailing sides of a button. This has potential to confuse the user. Use when representing an external link or new window icon to convey these behaviors to the user.
```
Rykus0, line 61: Should these be two separate points?
```suggestion
  - Avoid using icons on both the leading and trailing sides of a button. This has potential to confuse the user. 
  - Use when representing an external link or new window icon to convey these behaviors to the user.
```

website/docs/design/button.md
```diff
@@ -0,0 +1,135 @@
...
+  - Don't distract from nearby content
+  - Use for the lowest emphasis
+  - Use alone where content must be emphasized
+  - Use when multiple actions appear and a low-emphasis action is needed
+
+### Icon buttons
+
+- Take actions and make choices that are clearly indicated by a single icon
+- Circle
+  - Use when alone, as a floating action button or
```
Rykus0, line 45: Is the trailing 'or' intentional (connecting to the 'square') bullet?

website/docs/design/button.md
```diff
@@ -0,0 +1,135 @@
...
+- Action buttons should represent the main action in a button group. Do not use more than one action button in a group.
+- Use the same size button in groups. Do not use different sized buttons together in the same row.
+
+### Related patterns
+
+- Button vs. Link
+  - Use a link when you're navigating to another page or location. Examples include breadcrumb links, content links to other pages, and lists of links to related pages.
+    - Styles for links will deviate when placed in a global header and footer, or other navigation-type of element such as Tabs or Menus. This is to allow for a larger tap target and hierarchical presentation.
+  - Use buttons when performing an action, such as: "Submit," "Add a row," "Cancel," "Upload," etc.
+  - An action is almost always on the same page
```
Rykus0, line 76: I'm not sure what our standards are, but usually we want consistent punctuation within sets of lists.  There are a few other places like this where some items end with punctuation and others done.  My suggestion is to add periods to all of these.
```suggestion
  - An action is almost always on the same page.
```

website/docs/design/button.md
```diff
@@ -0,0 +1,135 @@
...
+  - Especially import in Danger sentiment buttons, to ensure the action being performed is clear.
+- Use concise, easy to scan, clear labels to indicate the action to the user
+- Do not use long, unnecessary phrases in button labels.
+- Use active voice to indicate the action
+- Do not use vague or generic labels such as "yes" or "No"
+
+## Accessibility
+
+- Expected actions for assistive technologies
+  - Ensure your label will make sense when spoken aloud with screen readers
```
Rykus0, line 102: We should either add to this, or in a separate bullet that the content should make sense independent of its context.  This is because a screen reader user may skip the context and jump to interactive content.

website/docs/design/button.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Alignment
+
+- Align buttons to the trailing edge of a container when using to prompt a user to navigate through a sequence, or when confirming an action (on the right for left-to-right languages and on the left for right-to-left languages).
+  - The highest emphasis action should be placed near the trailing edge of the container.
+- For focused tasks and single-page actions, align starting at the leading edge of the container.
+
+### Spacing and arrangement
+
+- Space buttons apart by `16px/1em`
```
Rykus0, line 131: Updating to rem since it is relative to the base font size, rather than the element's inherited font size, and more in line with the design intent (and implementation)

```suggestion
- Space buttons apart by `16px/1rem`
```

website/docs/design/button.md
bradryanbice: Ha, nope! I was thinking of another time when you'd use the circle shape. So far that's all I've got. I'll update. TY

website/docs/design/button.md
bradryanbice: This is correct. In some cases though, the bullets are not complete sentences so I am not using punctuation. I've reflected this in the content guidelines as well.

website/docs/design/button.md
bradryanbice: Good point, thanks! I'm teetering on if that portion is even necessary, but maybe for designer/developer communication and intent.

website/docs/design/button.md
bradryanbice: Good point. I'll add this as a separate bullet point.

caseybaggz (approved):

Issue #645, opened 2022-10-03
[Feature?]: Tooltip HS API


caseybaggz:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Add `getTooltipProps` and `getJSTooltipProps` to HS lib.

PR #666, opened 2022-10-06
Feat/645 tooltip css

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

Contributes to #645 

## What is the new behavior?

Adds Tooltip CSS API


packages/headless-styles/index.experimental.js
```diff
@@ -76,4 +76,6 @@ export { getJSTextareaProps } from './src/components/Textarea/textareaJS'
 export { getTextLinkProps } from './src/components/TextLink/textLinkCSS'
 export { getJSTextLinkProps } from './src/components/TextLink/textLinkJS'
 
+export { getTooltipProps } from './src/components/Tooltip/tooltipCSS'
```
caseybaggz, line 79: 😉 

```suggestion
export { getTooltipProps as unsafe_getTooltipProps } from './src/components/Tooltip/tooltipCSS'
```

packages/headless-styles/sandbox/src/components/Tooltip.jsx
```diff
@@ -0,0 +1,75 @@
+import { useState } from 'react'
+import { useEscToClose } from '@pluralsight/react-utils'
+import { getTooltipProps } from '../../../src'
+
+const positions = [
```
caseybaggz, line 5: Because this is such a large Array, I would honestly move it into a tooltipPositions.data.json file. Much more perfmant, cleaner, and reduces the risk of creating bugs in the JS file if you only need to update the positions Array.

packages/headless-styles/sandbox/src/components/Tooltip.jsx
```diff
@@ -0,0 +1,75 @@
...
+function TooltipEl(props) {
+  const [disabled, setDisabled] = useState(false)
+
+  const tooltipProps = getTooltipProps({
+    id: props.id,
+    position: props.position,
+    disabled: disabled,
+  })
+
+  function disable() {
```
caseybaggz, line 29: Anytime you pass a function into our helpers or a 3rd party component, you need to convert it to `useCallback`. Enable is fine because you are passing it into Raw JSX, but `disabled` for sure needs to use `useCallback`. Good pratice to remember.

packages/headless-styles/sandbox/src/components/Tooltip.jsx
```diff
@@ -0,0 +1,75 @@
...
+      </div>
+    </div>
+  )
+}
+
+export default function Tooltip() {
+  return (
+    <div id="tooltip">
+      <h2>Tooltips</h2>
+      <div
```
caseybaggz, line 55: We should just make a component that controls these styles so we can reuse it in the future. Take a little more advantage of the system Vite is giving us.

packages/headless-styles/src/components/Tooltip/tooltipCSS.module.css
```diff
@@ -0,0 +1,199 @@
...
+.bottomStartTooltip,
+.bottomEndTooltip {
+  padding-top: 12px;
+  top: 100%;
+}
+
+.bottomTooltip,
+.topTooltip {
+  left: 50%;
+  transform: translateX(-50%);
```
caseybaggz, line 83: This doesn't make sense to me. You are moving something 50% to the left...then reverting it on the line below with transform? There's also a [translate prop now](https://css-tricks.com/almanac/properties/t/translate/).

packages/headless-styles/src/components/Tooltip/tooltipCSS.module.css
```diff
@@ -0,0 +1,199 @@
...
+}
+
+.leftEndTooltip,
+.rightEndTooltip {
+  bottom: -4px;
+}
+
+/* Arrows */
+
+.tooltip::after {
```
caseybaggz, line 128: Is there a reason you stopped using composes overall? It's much cleaner vs using giantic blocks of re-selectors?

packages/headless-styles/src/components/Tooltip/tooltipCSS.ts
```diff
@@ -0,0 +1,55 @@
+// import { tooltip } from '@pluralsight/shared'
+import { createClassProp } from '../../utils/helpers'
+import {
+  createTooltipProps,
+  getDefaultTooltipOptions,
+  getTooltipClasses,
+} from './shared'
+import styles from './tooltipCSS.module.css'
+import type { TooltipOptions } from './types'
+const tooltip = true
```
caseybaggz, line 10: Did you mean to leave this hardcoded? You should have to use it if you've updated your brain with main?

packages/headless-styles/tests/tooltip/tooltipCSS.test.ts
```diff
@@ -0,0 +1,279 @@
+import { getTooltipProps } from '../../src'
+
+jest.mock('@pluralsight/shared', () => {
```
caseybaggz, line 3: No longer need this if you update your branch with main.

caseybaggz (commented):
Overall, since position is so CSS heavy, I would probably make another CSS file that puts all that logic there to keep the file smaller. Use modules to our advantage.

packages/headless-styles/index.experimental.js
Rykus0: I guess I misunderstood the new system.  I thought using the separate entry points meant we weren't prefixing anymore.

packages/headless-styles/src/components/Tooltip/tooltipCSS.module.css
Rykus0: The reason is that `left` positions relative to the parent's width, and `translate` positions relative to itself.  

So this puts the element's left edge at the center of the parent, then shifts the element back by half it's width to center it against the parent.

I wasn't aware of `translate` as a property.  I see the support is less than `transform` - is there an advantage to using this over transform? (Other than freeing it up for other things)  I can switch it, but don't want to break anything on older browsers if I don't have to.

packages/headless-styles/src/components/Tooltip/tooltipCSS.module.css
Rykus0: I guess mostly because I'm not used to using it and thinking that way.

packages/headless-styles/src/components/Tooltip/tooltipCSS.ts
Rykus0: Good catch, thanks.  I think I had tried updating and it didn't work.  I've had a lot of trouble working with the shared package, but that may have been before the rework you'd done, so I'll check again.

packages/headless-styles/src/components/Tooltip/tooltipCSS.module.css
caseybaggz: Well, fun fact - PS doesn't support older browsers. 😉 And, now that we have a bundler, we will be using tools like postCSS once we get our current plugins working so there's no need to be stuck in "old browser" mode. 

Also why we have this doc:
https://design.pluralsight.com/docs/next/development/getting-started/platforms

packages/headless-styles/src/components/Tooltip/tooltipCSS.module.css
caseybaggz: CSS modules baby, please keep using them so...

1. We are leveraging the tools provided by CSS modules
2. We stay consisten in the design of what is currently already done

packages/headless-styles/src/components/Tooltip/tooltipCSS.module.css
Rykus0: `translate` isn't supported by some of the versions listed in the platforms doc.  Should I still use it?  

My perspective is we should continue to use `translate`, since we don't have a need to switch and it has better support.

packages/headless-styles/src/components/Tooltip/tooltipCSS.module.css
Rykus0: As I'm updating, I'm realizing another reason I didn't, and that's because the relationship isn't 1:1, the mappings overlap.  So for example, `top`, `topStart`, and `topEnd` use _some_ of the same styles, then `top` and `bottom` share another set.  If I compose, then `bottom` will get styles it shouldn't.  Same for `topStart` and `topEnd`

caseybaggz: Right, so it's underlining a composition issue as a whole. So you should refactor the CSS so that you can use the composes in a more successful way. Does that make sense? You are overwriting styles right now, where composes is combining styles - totally different code design.

Rykus0: I shouldn't be overwriting anything right now, but I might be able to refactor to make a cleaner composition, though there will be some repetition where there was some reuse before.

Rykus0: If I use `composes` the way I have some of them now, then I _will_ need to overwrite certain properties.

packages/headless-styles/src/components/Tooltip/tooltipCSS.ts
Rykus0: Actually, we're just removing the flag logic entirely, right?  Since we're using the different entry points.

Issue #669, opened 2022-10-07
[Feature?]: Table API


caseybaggz:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

- [x] CSS
- [x] JS
- [ ] Svelte
- [x] SC
- [x] Chakra
- [x] MUI

PR #671, opened 2022-10-07
feat(shared): add table flag

caseybaggz:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

contributes #669 

## What is the new behavior?
Adds settings to support using `__EXPERIMENTAL__` flag value in project

### DIFF
--- a/packages/shared/src/featureFlags.ts
+++ b/packages/shared/src/featureFlags.ts
@@ -31,6 +31,8 @@ export const tabsHook = true
 // likely to include in an upcoming release.
 // -----------------------------------------------------------------------------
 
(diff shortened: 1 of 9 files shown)


packages/shared/src/featureFlags.ts
```diff
@@ -1,3 +1,5 @@
+declare const __EXPERIMENTAL__: boolean
```
caseybaggz, line 1: We have to use this here for build purposes because TS is too dumb to know to use the `global.d.ts` file...which is working everywhere else. Don't know, don't care.

Issue #692, opened 2022-10-18
[Feature?]: Select HS API


Rykus0:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

- [ ] CSS
- [ ] JS
- [ ] Svelte
- [ ] SC
- [ ] Chakra
- [ ] MUI

PR #696, opened 2022-10-19
Feat/692 select css

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

Contributes to #692 

## What is the new behavior?

Adds Select CSS API and tests


packages/headless-styles/sandbox/src/components/Select.jsx
```diff
@@ -0,0 +1,144 @@
...
+  getSelectProps,
+  getIconProps,
+  // getJSSelectProps,
+} from '../../../src'
+import selectOptions from '../data/options.data.json'
+
+function SelectField(props) {
+  const { onChange, ...options } = props
+  const { fieldOptions } = getFormControlProps(options)
+  const labelProps = getFormLabelProps({
```
caseybaggz, line 17: This would definitely be considered and "overloaded component" which is what we want to avoid building and promoting. Instead, let's break this up into single responsibility components that only manage a single thing. 

Makes the code easier to read, less bug risk, and maintain as well.

packages/headless-styles/src/components/Select/selectCSS.module.css
```diff
@@ -0,0 +1,99 @@
+.selectWrapper {
+  margin-top: 8px;
+  position: relative;
+  width: 100%;
+}
+
+.selectBase:not([multiple]) {
```
caseybaggz, line 7: Let's experiment using the CSS modules imports here to our benefit - see how it goes.

https://github.com/css-modules/css-modules#composing-from-other-files

packages/headless-styles/src/components/Select/shared.ts
```diff
@@ -0,0 +1,67 @@
+import { createA11yProps } from '../../utils/helpers'
+import type { Tech } from '../types'
+import type { SelectOptions, Size } from './types'
+
+const defaultSelectOptions = {
+  disabled: false,
+  describedBy: '',
```
caseybaggz, line 7: We should probably make a global defaultFieldOptions since we are doing that with the types, not sure why I've never thought of that until now... 🙈 😹

packages/headless-styles/sandbox/src/components/Select.jsx
Rykus0: Sure thing.  I copied this in from Input and thought about doing that, but never came back to it.

caseybaggz (approved):

PR #758, opened 2022-11-11
Feat/603 use submenu interaction

### LINKED ISSUE #603: [Feature?]: create useListKeyNavigation hook for menu interaction
### Duplicates

### Latest version

### Summary 💡

The hook (useListKeyNavigation) will control the focus management and keyboard events for menus, using a roving tabindex strategy.

### Examples 🌈

[Roving Tabindex](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Keyboard-navigable_JavaScript_widgets#technique_1_roving_tabindex)

[Keyboard interaction for Menu](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menu_role#keyboard_interactions)

### Motivation 🔦

_No response_

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

closes #603 

## What is the new behavior?

Adds hook for submenu interactions


packages/react-utils/src/hooks/menu/useSubmenuInteraction.ts
```diff
@@ -0,0 +1,183 @@
...
+        default:
+          null
+      }
+    },
+    [openMenuWithFocus, toggleMenu, toggleMenuWithFocus]
+  )
+
+  const handleMenuKeypress = useCallback(
+    (event: KeyboardEvent) => {
+      // eslint-disable-next-line default-case
```
caseybaggz, line 106: Oooh! You better not! Just add a default case that throws and error which says "Unknown key value passed to handleMenuKeypress." or something like that. 👍

packages/react-utils/tests/hooks/useSubmenuInteraction.test.tsx
```diff
@@ -0,0 +1,254 @@
+import { type ReactNode } from 'react'
+import { render, screen, userEvent } from 'test-utils'
+import { useSubmenuInteraction } from '../../src'
+
+type MenuProps = {
```
caseybaggz, line 5: This should use an `interface` which extends `ReactNodeWithChildren` or something like that type. Then you don't need to include Children. 😉

Rykus0: Thanks!

packages/react-utils/src/hooks/menu/useSubmenuInteraction.ts
Rykus0: I didn't want to do that because other keystrokes are valid, we're just not handling them.  I had a default that just called a noop, but didn't like that so I changed it.

caseybaggz: Well, not including it adds a higher risk of bugs which is why there is an eslint rule for it. So, I would much rather have a blank default case `default: break;` than increase risk by using a eslint disable comment if that makes sense?

Rykus0: Yeah, definitely.  I thought I'd tried that and it didn't allow it, but when I tried it again now it was fine so I must've done something else.

caseybaggz (approved):

PR #779, opened 2022-11-15
docs: update migration.md to match latest standards

caseybaggz:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

Migration page had outdated information.

## What is the new behavior?
Updates the page with the latest examples and recommendations via the Beta API.

## Other information
Click the Netlify preview link to see a PR build of the docs.

### DIFF
--- a/website/docs/development/getting-started/migration.md
+++ b/website/docs/development/getting-started/migration.md
@@ -1,5 +1,5 @@
 ---
-tags: [Development, Getting Started, Migration, PSDS Classic]
+tags: [PSDS Classic]
 title: Migration
 ---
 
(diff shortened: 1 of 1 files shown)


website/docs/development/getting-started/migration.md
```diff
@@ -71,99 +77,191 @@ In order to use PD, we need to get it into your project first. For most projects
...
 import { getButtonProps } from '@pluralsight/headless-styles'
 import type { ButtonOptions } from '@pluralsight/headless-styles/types'
 
-// 💡 Keeping logic out of your components improves performance
 const cancelBtnProps = getButtonProps({
   sentiment: 'danger',
 })
 
-export default function CancelButton(props: ButtonOptions) {
+function CancelButton(props: ButtonOptions) {
```
Rykus0, line 111: Should this type be their props, and not the ones for Headless-styles?

caseybaggz: @Rykus0 in this example, the point is showing we provide a props type. We don't want to assume all people either have props or are even using TS for that matter. Make sense?

Rykus0: Yeah, I was thinking that the type defines the options for our API (sentiment, etc), and not the props that would be passed to the component (onClick, etc).  I see that the ButtonOptions type also includes the base props for Button, so I think it works in this case, but I don't think we did that everywhere.

Rykus0 (approved):

website/docs/development/getting-started/migration.md
caseybaggz: Nope! Only for the Button API since it matches the use-case more.

PR #809, opened 2022-11-18
Feat/713 popover

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

contributes to #713 

## What is the new behavior?

Adds CSS Popover API & tests
Refactors a utility helper for generating aria labeling attributes


packages/headless-styles/src/components/Popover/popoverCSS.module.css
```diff
@@ -0,0 +1,98 @@
+@keyframes fadeIn {
```
caseybaggz, line 1: I feel like we don't need half of this because it already exists in Tooltip? Why aren't we just using those styles with the small additions we need for Popover?

packages/headless-styles/src/components/Popover/popoverCSS.ts
```diff
@@ -0,0 +1,78 @@
+import { popover } from '@pluralsight/shared'
+import { createClassProp } from '../../utils/helpers'
+import {
+  createPopoverProps,
+  getDefaultPopoverOptions,
+  getPopoverClasses,
+} from './shared'
+import popoverStyles from './popoverCSS.module.css'
+import positionStyles from './popoverPositioning.module.css'
+import pointerPositionStyles from './pointerPositioning.module.css'
```
caseybaggz, line 10: Is there a reason we duplicated these and aren't just using the Tooltip version?

packages/headless-styles/src/utils/a11yHelpers.ts
```diff
@@ -0,0 +1,18 @@
+import { warning } from '@pluralsight/shared'
+
+export function getAccessibleLabelProps(labelId?: string, ariaLabel?: string) {
+  warning(
```
caseybaggz, line 4: Let's kill this. We shouldn't need it because we are using TS.

caseybaggz: After looking at this function, I think it's too vague because you are returning 2 different types of label props which causes the params to get shady.

So, you either need to split them into 2 functions `getA11yLabeledBy` & `getA11yLabel` or change the function purpose in general. Right now from a reading perspective, it just doesn't really make sense why it can't be done statically in the source?

Also, for utility functions, you should always use the a11y naming vs the long accessible. 👍

PR #819, opened 2022-11-21
Narrow types for `createJSProps` and `createCSSObj` return values

josh-degraw:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes
- [x] Other... Please describe: Typescript changes

In testing out headless styles, I found it difficult to know what styles were available to reference, because the `styles` entry on all of the `getJS*Props` functions is typed as `Record<string, unknown>`.

## What is the new behavior?

If these changes are accepted, it should improve type definitions on `getJS*Props` so that rather than just having all `styles` entries be `Record<string, unknown>`, it would use the actual values given and infer the types and show what selectors are actually used.

### DIFF
--- a/packages/headless-styles/src/utils/helpers.ts
+++ b/packages/headless-styles/src/utils/helpers.ts
@@ -76,7 +76,10 @@ export function createSvelteObj(classname = '') {
   return { class: classname }
 }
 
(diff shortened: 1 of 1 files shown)


caseybaggz (approved):
Great addition @josh-degraw !

PR #842, opened 2022-11-28
feat(headless-styles): make createJSProps generic

josh-degraw:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes
- [x] Other... Please describe: Typescript updates


caseybaggz:
@josh-degraw looks like we are getting some type errors in the failing check. If you click "details" you can see what is going on.

packages/headless-styles/src/utils/helpers.ts
```diff
@@ -12,6 +12,22 @@ export type CSSCustomSelectors = Record<string, CSS.Properties>
 export type CSSPseudos = Record<CSS.Pseudos, CSS.Properties>
 
 export interface CSSObj extends CSS.Properties, CSSPseudos {}
+
+// Helper type to expand and simplify computed types
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type Simplify<T> = T extends any ? { [K in keyof T]: T[K] } : never
```
caseybaggz, line 18: We can't ship the `any` type which is why we have the rule as it will break many teams build's who no longer use that type (TS recommends using `unknown` instead).

packages/headless-styles/src/utils/helpers.ts
```diff
@@ -12,6 +12,22 @@ export type CSSCustomSelectors = Record<string, CSS.Properties>
...
+
+type CorrectCSSTypes<T> = Simplify<{
+  // Iterate through the keys of the object
```
caseybaggz, line 21: Let's kill all the comments. We normally only keep those for code that is not readable which this is very readable. 😄

caseybaggz (commented):
Build checks are failing due to type errors.

packages/headless-styles/src/utils/helpers.ts
josh-degraw: That's a good rule but for what it's worth, no `any`s would be shipped if this change was used, all this type does is essentially generate a new, easier-to-read type based on the `T`, i.e. it flattens the unions/intersection types into a single type. The `extends any` part just tells typescript to distribute the properties of the given type.

However, this might be related to the build failure on the PR, and it's not strictly necessary, it just made seeing the type information via mouse-over work better, so I'm going to try pulling that out and seeing if that might help fix the build issue.

caseybaggz:
Outdated. Fixed in #847

josh-degraw:
> Outdated. Fixed in #847 

That PR actually did not fix the issue described here

caseybaggz:
@josh-degraw looks like there are some lint errors. I think the TS one will probably be the easiest to "fix them all".

packages/headless-styles/src/utils/helpers.ts
```diff
@@ -130,10 +131,10 @@ export function createClassProp(tech: Tech, classes: ClassOptions) {
   return createCSSObj(classes.defaultClass)
 }
 
-export function createJSProps(styles: GeneratedStyles) {
+export function createJSProps<Styles extends GeneratedStyles>(styles: Styles) {
   return {
     cssProps: transformStyles(styles),
-    styles: styles as unknown as CSSObj,
+    styles,
```
caseybaggz, line 137: By removing the CSSObj type is causing failures on all the tests because we are removing the property typing. We need to figure out a way to keep that in so TS doesn't "break the world".

packages/headless-styles/sandbox/plugins/css-to-ts/convert-inputs/write-file.js
```diff
@@ -26,7 +26,7 @@ const writeToFile = (outPath, input, outputType) => {
     }
 
     Object.entries(input.contents).forEach((key) => {
-      const body = indentObject(key[1], 2)
+      const body = indentObject(key[1], 2) + ' as const'
       writeFileSync(path.join(fileOutPath, `${key[0]}.ts`), 'export default ')
```
caseybaggz, line 30: I think we can enhance this to better achieve what is trying to be done here. We can instead change line 30 to do something like (not perfect, just an example):

```
      const styleKey = `${keyWithoutModule]Styles`
      writeFileSync(path.join(fileOutPath, `${key[0]}.ts`), `export const ${styleKey} = `)
```

.gitignore
```diff
@@ -25,4 +25,10 @@ packages/headless-styles/npm/types
 packages/react-utils/npm/types
 
 # Mac system files
-.DS_Store
\ No newline at end of file
+.DS_Store
+
+*node_modules
+.vscode/
```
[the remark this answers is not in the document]
josh-degraw: My VSCode just added another file under `.vscode` so I just updated it to ignore anything that wasn't already in there. The vscode settings were already included in the repo so I excluded them here just to not break anything.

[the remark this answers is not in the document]
josh-degraw: Along the same lines, is `.yarn/cache` intentionally included in the repo? It makes modifications any time I install the dependencies and I doubt you'd want me or any other contributer to unintentionally check in a bunch of changes to the yarn cache haha.

packages/headless-styles/src/utils/helpers.ts
josh-degraw: Yeah, I'm aware of that, this is the core of what I'm trying to improve with this PR.
I had originally added a utility type that dynamically remapped the keys without needing to know the constant values, but I think I just figured out a way to accomplish what I wanted without needing to modify the build output and that doesn't break the tests.

I added a type that essentially just intersects any nested selectors with a `CSSObj` so those nested selectors can be seen in type hints etc. I'm not sure it's perfect I think it'll accomplish what I'm hoping for in a bit of a simpler way.

caseybaggz:
Going to close this since it is pretty stale.

Issue #675, opened 2022-10-12
[Feature?]: transformCSStoTS to support CSS Modules


Rykus0:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Replace `composes` properties with the contents of the referenced class(es).
Other properties within the rule should override any from the composed class(es).
Multiple classes should be applied sequentially, with the latter overwriting the former.

**Should fully support all CSS Modules features. Should make it a new plugin probably.**

### Examples 🌈

Given:
```css
.classA {
  border: 1px solid;
  color: red;
  padding: .5em;
}

.classB {
  background-color: white;
  padding: 1rem;
}

.classC {
  composes: classA classB;
  color: blue;
}
```

`.classC` will become
```js
classC: {
  border: '1px solid',
  backgroundColor: 'white',
  color: 'blue',
  padding: '1rem',
}
```

### Motivation 🔦

Currently in the JS APIs, we are manually assigning the composed classes.  Allowing the tool to handle this will reduce the size of our code, as well as opportunity for errors.  It will also improve the consistency between the CSS and JS API outputs, and reduce maintenance by consolidating control.

PR #864, opened 2022-12-02
Feat/compose css in js

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

contributes to #675 

## What is the new behavior?

Adds support for composing local class names
Re-generates style objects and accordingly updates JS APIs 
Adjusted application of styles for Progress to simplify and avoid some conflicts/reduce potential bugs
Fixed composition of classes in Badge causing background-color issues in some cases
Checkbox & Radio - resolved a bug where direction was not properly applied in JS API


packages/headless-styles/sandbox/src/components/CircularProgress.jsx
```diff
@@ -22,9 +23,11 @@ function CircleProgress(props) {
...
 
 export default function CircularProgress(props) {
-  if (props.logJS) {
-    console.log({ ...getJSCircularProgressProps({ now: 32 }) })
-  }
+  useEffect(() => {
+    if (props.logJS) {
+      console.log({ ...getJSCircularProgressProps({ now: 32 }) })
+    }
+  })
```
caseybaggz, line 30:
```suggestion
  }, [])
```

packages/headless-styles/src/components/Admonition/generated/admonitionCSS.module.ts
```diff
@@ -35,10 +35,16 @@ export default {
   admonitionDescription: {
     composes: 'admonitionText',
```
caseybaggz, line 36: I think if we do this, we should also kill the `composes` prop since it's a generated output.

caseybaggz: Likewise, the composes result should technically be first in the Object vs. last.

packages/headless-styles/src/components/Badge/badgeCSS.module.css
```diff
@@ -37,11 +37,11 @@
 /* sentiment */
 
 .defaultBadge {
-  composes: filledBadge;
+  composes: baseBadge;
```
caseybaggz, line 40: This will break the styles because `filledBadge` adds additional styles.

packages/headless-styles/src/components/Popover/popoverJS.ts
```diff
@@ -35,15 +44,19 @@ export function getJSPopoverProps(options?: PopoverOptions) {
...
+      ...tooltipStyles.tooltipContentBase,
+      ...popoverContentStyles,
+      ...popoverContentPositionStyles,
       ['&::after']: {
-        ...tooltipStyles.tooltipContentBase['&::after'],
-        ...styles.popoverContent['&::after'],
-        ...contentPositionStyles['&::after'],
+        ...(popoverContentStyles[
+          '&::after' as keyof typeof popoverContentStyles
+        ] as Record<string, string>),
```
caseybaggz, line 56: This is not the right style override, you need to use the CSSObj we have in helpers.

packages/headless-styles/src/components/Progress/generated/progressCSS.module.ts
```diff
@@ -29,4 +22,10 @@ export default {
   inset: {
     borderRadius: 'initial',
   },
+  sSize: {
+    height: '0.5rem',
```
caseybaggz, line 26: Looks like these are missing the composes styles.

packages/headless-styles/src/components/Progress/progressCSS.module.css
```diff
@@ -27,6 +22,10 @@
   border-radius: initial;
 }
 
+.sSize {
+  height: 0.5rem;
```
caseybaggz, line 26: These require composes.

packages/headless-styles/src/components/Admonition/generated/admonitionCSS.module.ts
Rykus0: Yeah, I had this on my list and was going to get it when I did the external composes, but I'll just do it now so it's done.

Rykus0: I'd also considered sorting the properties, since they're out of order when merged.  Since we're removing composes, I won't worry about that one and maybe we can still sort them later if we feel the need.

packages/headless-styles/src/components/Badge/badgeCSS.module.css
Rykus0: It was breaking the styles to leave it in because `filledBadge` has a background color that overwrote the one for `outlineBadge`.  There's probably a better way to handle it, though.  We're applying sentiment and usage largely exclusively, but it's possible to apply them together, so there are also some overlap issues.  With both separate options inheriting from the same base, that base would need to exclude anything that is unique to either of those sets of styles.

I'll work on it and see what I can do to make it cleaner.

packages/headless-styles/src/components/Progress/generated/progressCSS.module.ts
Rykus0: I had to take it out because of the way they were being applied originally, but I can probably add it back in now that it's only used on the wrapper.

packages/headless-styles/src/components/Badge/badgeCSS.module.css
Rykus0: I think it works in the CSS API because of the source order of the classes

caseybaggz: @Rykus0 you think or you know? You should be validating the components still match the UI Kit in the sandbox with this level of changes. 😉

packages/headless-styles/sandbox/plugins/css-to-ts/convert-rules/convertToJS.js
```diff
@@ -42,13 +44,25 @@ const convertRules = (rules, res = {}) => {
...
       // Convert standard CSS rules
       const standardProp = standard(rule, result)
+
       Object.entries(standardProp).forEach(([key, value]) => {
-        const sanitizedValue = findAndReplaceVars(value)
+        const { composes, ...sanitizedValue } = findAndReplaceVars(value)
+
+        if (composes) {
+          toBeComposed[key] = composes
+          delete result[key].composes
```
caseybaggz, line 53: We should be using immutability here which is both safer and faster.

caseybaggz:
Too risky of the moment. Will pause and come back to the feature.

PR #930, opened 2022-12-21
Expansion

### LINKED ISSUE #675: [Feature?]: transformCSStoTS to support CSS Modules

josh-degraw:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

Generated CSS.ts files don't currently apply CSS module composition. 

## What is the new behavior?

The relevant CSS is applied during the `start:sandbox` process. Classes imported from other files are explicitly imported to allow for the script to not care about load order.

## Other information
Closes #675


caseybaggz:
@josh-degraw looks like there are a couple of conflicts that need resolving. Once those are fixed and pushed you can now validate your changes in the Codesandbox links (I guess CS just needed 2023 to start working again...Lol.).

caseybaggz:
@josh-degraw looks like there are a couple of conflicts that need resolving. Once those are fixed and pushed you can now validate your changes in the Codesandbox links (I guess CS just needed 2023 to start working again...Lol.).

packages/headless-styles/sandbox/plugins/css-to-ts/convert-rules/convertToJS.js
```diff
@@ -52,12 +52,39 @@ const convertRules = (rules, res = {}) => {
...
+      } else if (composes.includes(' from ')) {
+        // Ignore here, will be appended in later step
+        Object.assign(result[className], value)
+      } else {
+        composes.split(' ').forEach((otherKey) => {
+          Object.assign(result[className], obj[otherKey])
+        })
+      }
+    }
+    Object.assign(result[className], overrides)
```
Rykus0, line 76: This will overwrite nested properties, rather than merging them (see tooltipPosition styles with `&::after`), so some will be lost.

packages/headless-styles/sandbox/plugins/css-to-ts/convert-rules/convertToJS.js
```diff
@@ -52,12 +52,39 @@ const convertRules = (rules, res = {}) => {
...
+    const { composes, ...overrides } = value
+    result[className] = {}
+    if (composes) {
+      if (composes in obj) {
+        Object.assign(result[className], obj[composes])
+      } else if (composes.includes(' from ')) {
+        // Ignore here, will be appended in later step
+        Object.assign(result[className], value)
+      } else {
+        composes.split(' ').forEach((otherKey) => {
```
Rykus0, line 71: I would split on 1 or more whitespace characters, rather than a single space character, just in case.

```suggestion
        composes.split(/\s+/).forEach((otherKey) => {
```

packages/headless-styles/sandbox/plugins/css-to-ts/convert-inputs/write-file.js
```diff
@@ -18,40 +19,87 @@ const writeToFile = (outPath, input, outputType) => {
...
+
+      let importName
+      if (generatedImportFileName in imports) {
+        importName = imports[generatedImportFileName]
+      } else {
+        importName = file.match(moduleReg)[3]
+        imports[generatedImportFileName] = importName
+      }
+
+      output += `...${importName}.${name},`
```
Rykus0, line 90: This will result in common nested properties being overwritten

packages/headless-styles/sandbox/plugins/css-to-ts/convert-inputs/write-file.js
```diff
@@ -18,40 +19,87 @@ const writeToFile = (outPath, input, outputType) => {
...
+function getImports(str) {
+  const imports = {}
+  let output = ''
+
+  for (const line of str.split('\n')) {
+    const match = line.match(/"?composes"?: "(.+) from '(.+)'"/)
+
+    if (match) {
+      const [, name, file] = match
+      const moduleReg = /(..\/)+(.+)\/(.+)\.module\.css/
```
Rykus0, line 75: My suggestion is assuming the first two dots are intended as the dot character, and not any character in order to match "../(path)/(file)"

This will break if we reference any files in a different path (more likely to be `./` than anything else), in which case, we can make one of the dots optional. I don't think we'll need flexibility beyond that

```suggestion
      const moduleReg = /(\.?\.\/)+(.+)\/(.+)\.module\.css/
```

Rykus0 (changes requested):
Thanks, Josh, this is great!  There are a couple places where we're applying the composed properties that they can overwrite nested properties, rather than merging them.  So we'll need to address those.

caseybaggz:
@josh-degraw something seems to be dramatically breaking the CI pretty hardcore throughout the history of this PR which is not a good sign. I'm also seeing a lot of extra work thrown in that doesn't relate to the issue. 

So, I'm going to start a branch that cherry-picks the non-CSS module updates and once we get that merged, I'm wondering if it might be beneficial for you to start a new PR based on that branch state updating only the CSS module stuff and see what the CI does?

I would love to get this work in before we cut the RC release next week, so both @Rykus0 and I will pair with you as much as it takes for us to try and get it in. Thanks so much for the strong foundation and great work! Hopefully, together, we can get this shipped successfully. 👍

packages/headless-styles/sandbox/plugins/css-to-ts/convert-inputs/write-file.js
```diff
@@ -18,41 +19,149 @@ const writeToFile = (outPath, input, outputType) => {
...
+  return {
+    imports: imports + '\n',
+    output,
+  }
+}
+
+/**
+ * @param {object} value
+ * @param {Set<string,string>} importMap
+ */
```
Rykus0, line 90: We don't usually use jsdoc comments like this, so you can remove them.  (plus they tend to quickly lose accuracy, which you can see here this one specifying `Set` instead of `Map`

josh-degraw: Yeah good catch, I just put those in for me to keep track while developing and just forgot to remove them 🤦🏻‍♂️

josh-degraw:
> @josh-degraw something seems to be dramatically breaking the CI pretty hardcore throughout the history of this PR which is not a good sign. I'm also seeing a lot of extra work thrown in that doesn't relate to the issue.

I had to update some dependencies to handle Typescript 4.9 which it looks like causes all of the CI steps to fail: 
```sh
> The lockfile would have been modified by this install, which is explicitly forbidden.
```

caseybaggz:
> `> @josh-degraw something seems to be dramatically breaking the CI pretty hardcore throughout the history of this PR which is not a good sign.
> 
> I had to update some dependencies to handle Typescript 4.9 which it looks like causes all of the CI steps to fail:
> 
> ```shell
> > The lockfile would have been modified by this install, which is explicitly forbidden.
> ```
> 
> I pulled that and the related changes out into a [separate branch](https://github.com/josh-degraw/tva/pull/1) and now I'm just getting a few failing tests, which is super fun because I can't seem to run the tests locally due to (I assume) the pnp configuration and windows not getting along 🙃

Hrm...let me look into this today. I'm about to post a PR that I'll add you for review in that updates all non-issue updates you have in here so we can keep the PR cleaner regarding the issue it resolves. Hopefully, it should help the review as well of this one since by nature it is so large.

caseybaggz:
> ```shell
> The lockfile would have been modified by this install, which is explicitly forbidden.
> ```

@josh-degraw can you verify which Yarn version you are running locally within the repo?

josh-degraw:
> > ```shell
> > The lockfile would have been modified by this install, which is explicitly forbidden.
> > ```
> 
> @josh-degraw can you verify which Yarn version you are running locally within the repo?

Running locally shows 3.3.1, but that message above was from the CI job FWIW

josh-degraw:
> Ah, but we need to do it. This week is the perfect time to knock it out. Technically, with Testing Lib, we aren't supposed to have to do it...but...CI's be CI's.

Haha for sure, I just meant that what I did was actually incorrect and broke a few tests, so it's probably best to just be its own separate PR haha

caseybaggz:
@josh-degraw review officially open for #983 which I tagged you in. Not much, but should lighten the load a little in here.

.gitignore
```diff
@@ -16,6 +16,9 @@ temp
 packages/icons/npm
 packages/design-tokens/npm
 
+node_modules/
+*/**/node_modules/
```
caseybaggz, line 20: We don't need this one.

```suggestion
*/**/node_modules/
```

packages/headless-styles/sandbox/plugins/css-to-ts/convert-rules/convertToJS.js
```diff
@@ -57,10 +57,48 @@ const convertToJS = (input) => {
...
-    return convertRules(rules)
+    const initial = convertRules(rules)
+
+    return applyCompositions(initial)
   } catch (err) {
     throw new Error(`Invalid CSS input: ${err}`)
   }
 }
 
 export default convertToJS
```
caseybaggz, line 68: Let's throw the `export default` on the convertToJS line so we don't have it floating radomly in the middle of the file.

Rykus0:
@caseybaggz Looks like only design-tokens built in the sandboxes

packages/headless-styles/src/components/Badge/badgeJS.ts
```diff
@@ -24,12 +24,28 @@ export function getJSBadgeProps(options?: BadgeOptions) {
...
   const props = createBadgeProps(defaultOptions)
   const { sentimentClass, sizeClass, usageClass } =
-    createBadgeClasses<typeof styles>(defaultOptions)
+    createBadgeClasses(defaultOptions)
+
+  const uniqueOptions = Boolean(options?.sentiment) !== Boolean(options?.usage)
+
+  let sentimentAndOrUsage
+
+  if (uniqueOptions) {
```
caseybaggz, line 33: Let's make this a function outside of this helper that is the value of the variable`sentimentAndOrUsage`. Doing this will eliminated the floating condition in the middle of this helper to keep it's responsibility clean and well as keep the helper clean.

packages/headless-styles/src/utils/helpers.ts
```diff
@@ -76,9 +79,25 @@ export function transformStyles(styleObject: NestedGeneratedStyles) {
...
       ${prev.trim()}
-      ${propName} ${transformValue(currentValue as NestedStyleValue)}
+      ${propName} ${transformValue(currentValue)}
     `
     }, '')
     .trim()
     .replace(/^ {2,12}/gm, '') as unknown as TemplateStringsArray
 }
+
+type ExtractedObject<
```
caseybaggz, line 89: Let's go ahead and throw this in the helpers/types file as well

packages/headless-styles/src/utils/types.ts
```diff
@@ -58,6 +57,8 @@ export type NestedStyleObject =
   | Record<`--${string}`, string>
   | CSSKeyframes
 
+export type GeneratedStyles = Record<string, NestedGeneratedStyles>
```
caseybaggz, line 60: Let's move this up a few lines to keep this list in alphabetical order so contributors can glance and find faster.

caseybaggz (changes requested):
Great stuff! Just need to fine tune a few things, then test in the sandboxes.

caseybaggz:
> @caseybaggz Looks like only design-tokens built in the sandboxes

Interesting. @josh-degraw after you get the latest review comments in and pushed up, can you double check your fork is updated? I'm noticing the Codesandbox list is our old one and not the current and since this PR has a big code footprint it is critical we test hardcore in the sandbox.

.gitignore
josh-degraw: Yeah I guess you're right, when I had to run it locally without pnp it added node modules to each packages folder, but I _think_ the pnp stuff might finally be mostly working for me know so it's probably fine

.gitignore
```diff
@@ -16,6 +16,9 @@ temp
 packages/icons/npm
 packages/design-tokens/npm
```
josh-degraw, line 17: I noticed you also removed the exclusion of the `.vscode` directory from the `.gitignore` I have some extensions that add files in there and sometimes I like to add some scripts to a `tasks.json` file that I don't know if makes sense to include in the repo, so I put that in there. If you want to ignore any other potential files that might be added there but keep the ones that are already there, we could add these:

```.gitignore
.vscode/*
!.vscode/extensions.json
!.vscode/launch.json
!.vscode/settings.json
```

.gitignore
caseybaggz: Yeah, if pnp doesn't work on a local instance - that needs to be corrected over changing the repo for the preferred scenarios where it is expected work.

josh-degraw: I noticed you also removed the exclusion of the `.vscode` directory from the `.gitignore`. I put that in there because I have some extensions that add some files in there and also in different projects sometimes I like to add some scripts to a `tasks.json` file that I don't think should be included in the repo. If you want to ignore any other potential files that might be added there but keep the ones that are already there, we could add these:

```.gitignore
.vscode/*
!.vscode/extensions.json
!.vscode/launch.json
!.vscode/settings.json
```

.prettierrc.json
```diff
@@ -1,4 +1,5 @@
 {
   "semi": false,
-  "singleQuote": true
-}
+  "singleQuote": true,
+  "parser": "typescript"
```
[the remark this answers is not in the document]
josh-degraw: Yeah, this was put in there because I added prettier inline usage in the css generation script, but I added this line because of a different issue in an unrelated change that I was trying out. Reverted this.

josh-degraw:
How can I see the sandbox based on my latest changes? The only links I'm seeing are from like a month ago

caseybaggz:
> How can I see the sandbox based on my latest changes? The only links I'm seeing are from like a month ago

Hm. I think there is possibly a complex issue with the Codesandbox CI here that ultimately stems from the changes in the repo for the past month:

1. Name change
2. Codesandbox re-auth with new name

So, I'm wondering if we need to close this PR and open a new one to trigger the current Codesandbox auth CI? 🤔

josh-degraw:
> I'm wondering if we need to close this PR and open a new one to trigger the current Codesandbox auth CI? 🤔 
I tried that in #930, even put the branch directly in this repo, and I'm not seeing _any_ sandbox links there? Am I just missing something?

caseybaggz (approved):
Will have to test reactively once merged and `next` build has shipped in sandbox of choice.

packages/headless-styles/src/components/Tooltip/generated/tooltipPositioning.module.ts
```diff
@@ -12,6 +12,9 @@ export default {
     '&::after': {
       bottom: '0',
       transform: 'translateY(0.5em) rotateZ(-45deg)',
+      left: '1em',
+      translate: '-50% 0',
+      right: '0.5rem',
```
Rykus0, line 17: Do you know where these might be coming from?  Think it might be a bug.  In the meantime I'm going to investigate and I'll let you know if I find anything.

Rykus0: Ah.  Looks like it's merging based on the `composes` rule in `.topContent` and applying that here as well.  
My first guess is that the original object is being merged into, rather than a copy.

packages/headless-styles/sandbox/plugins/css-to-ts/convert-rules/convertToJS.js
```diff
@@ -52,15 +52,51 @@ const convertRules = (rules, res = {}) => {
...
+}
+
+function deepMerge(source, target) {
+  for (const [key, value] of Object.entries(target)) {
+    // Overwrite primitive values, merge objects together
+    if (key in source && typeof value === 'object') {
+      deepMerge(source[key], value)
+    } else {
+      source[key] = value
+    }
```
Rykus0, line 100: @josh-degraw  found it.  This is causing the bug I saw in tooltipPositioning.  If `key` is not in `source` and `value` is an object, we are directly assigning `value` (pointer to the object), causing it to also change when `source` changes again later.

josh-degraw: Oof. Good catch! Got a fix incoming

Rykus0:
Could you also add `  "parser": "babel"` to `prettierrc.json`?  That will resolve the warnings we're getting when running the dev sandbox.

josh-degraw:
> @josh-degraw Could you also add ` "parser": "babel"` to `prettierrc.json`? That will resolve the warnings we're getting when running the dev sandbox. And I think that's the last thing for me.

Just needed to update the inline usage to use the "typescript" parser. No need to mess with the global config after all

Rykus0 (approved):

caseybaggz (approved):

caseybaggz:
@josh-degraw @Rykus0 about to merge this. Once the main branch CI jobs complete, please startup a sandbox and test out as many JS APIs as possible to make sure we are not shipping any recursive bugs.

Massive work @josh-degraw ! Thanks again for this huge feature. 👏 🙌 🎉

josh-degraw:
> Massive work @josh-degraw ! Thanks again for this huge feature. 👏 🙌 🎉

Glad to help! This was a fun problem to solve :)

caseybaggz:
PR will be followed up by #992 for validation in the dev sandbox.

Issue #868, opened 2022-12-05
[Feature?]: add class merge in HS


caseybaggz:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Currently, HS doesn't offer any `class` utilities which forces the user to do custom logic to combine class styles  which can be tedious in certain scenarios (i.e. user also uses CSS Modules in combination with HS CSS modules). It would be nice to be able to just extend the classes to allow the native CSS API to be more fluid.

### Examples 🌈

```javascript
const oneColItem = getGridItemProps({ colSpan: 1, class: styles.container })
```

results in:
```typescript
interface GridItemProps {
  className: 'ps-grid-item user_cssModule_class ps_gen_cssModule'
  style: ...
}
```

Rykus0:
This gets complicated for components with multiple elements - which element should the classes go on?

We could export a helper to merge classes with a specific set of returned props, but that could get awkward.  

For now, I will add this only to the helpers which apply to a single element.  For those with multiple, I will apply the classes to the outer-most element.

Rykus0:
I was thinking about refactoring `createClassNameProp` to accept an array instead of the two strings.  That would allow us to skip any undefined classes and I feel is easier to work with and read. 

Usage would look something like this:

```
function getComponentProps(options) {
	 return {
		 ...createClassNameProp([
			'ps-switch-container',
			styles.container,
			styles[sentimentClass],
			defaultOptions.className,
		 ])
	 }
}
```

Rykus0:
For multi-element components, we could accept an object whose keys align with the helper's output and assign the classnames that way. 

```
getComponentProps({
	className: {
		 wrapper: 'class-name',
		 element: 'element-class',
	}
})
```

caseybaggz:
Going to close this since we are not doing it any time soon.

PR #939, opened 2023-01-03
Feat/868 class merge single

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

contributes to #868 

## What is the new behavior?
Adds a parameter to the className prop helper for a user-provided class and, if present, merges it with the existing list.
Applied this to single-element components.  

## Other information
Multi-element components will need a more complicated approach and some discussion over how we want to handle it (I have a few ideas).
Considering a refactor of `createClassNameProp` to accept an array instead

### DIFF
--- a/packages/headless-styles/src/components/Badge/badgeCSS.ts
+++ b/packages/headless-styles/src/components/Badge/badgeCSS.ts
@@ -34,7 +34,8 @@ export function getBadgeProps(options?: BadgeOptions) {
     badge: {
       ...props.badge,
       ...createClassNameProp(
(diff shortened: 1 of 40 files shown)


caseybaggz:
@Rykus0 let's hold off on this one. We are not sure we absolutely need it vs. other issues like the list update. This is definitely a nice to have that has a very, very large code footprint that we shouldn't ship at this moment unless we are 100% sure we need it.

PR #941, opened 2023-01-03
build(headless-styles): make headless-styles tree-shakeable

### LINKED ISSUE #940: [Bug?]: headless-styles isn't tree-shakeable
### Duplicates

### Latest version

### Current behavior 😯

See https://github.com/ThatJoeMoore/ps-tva-building for example build code.

When building a simple bundle which only uses Button and FieldMessage, the entire headless-styles package gets included in the output.

Input:

```js
import { getButtonProps, getFieldMessageProps, getJSButtonProps, getJSFieldMessageProps } from "@pluralsight/headless-styles"

console.log({
  css: {
    button: getButtonProps(),
    fieldMessage: getFieldMessageProps()
  },
  js: {
    button: getJSButtonProps(),
    fieldMessage: getJSFieldMessageProps()
  }
})
```

Minified output sizes built with Webpack, Rollup, and esbuild:

```
251K headless-styles-esbuild.js
253K headless-styles-rollup.js
158K headless-styles-webpack.js
```

If you inspect the output bundles, the CSS for all of the components is included, unminified, and with inline source maps.

### Expected behavior 🤔

Only the CSS/JS that is actually used should be included in the output bundle.

### Steps to reproduce 🕹

Steps:

1. Clone https://github.com/ThatJoeMoore/ps-tva-building
2. Run `yarn build:minify` to get minified bundles. Or, run `yarn build` to get the unminified, easer-to-inspect version.

### Context 🔦

Having worked with a similar design system package in the past, I've got some recommendations (which I will submit as a PR shortly).

Most of these recommendations basically boil down to the differences between a library and an application, and work on the assumption that, if a consumer wants something like minification, it'll take care of it at that level. Often times 'production' app-style optimizations have negative impacts in a library, as it prevents the consumer from making smarter, use-case specific optimizations (like tree-shaking).

1) By outputting a single output file with all of the components, downstream projects cannot tree-shake the component CSS, as the `style-inject` pieces are not side-effect free. This can be resolved by using `preserveModules: true` in the rollup config, as this keeps the files separate, and the `sideEffects: false` setting in `package.json` works properly. Basically, the tree-shakers will keep EVERYTHING that has side effects inside a file that gets anything used, as they can't tell which parts are used for what, so splitting up the files allows the tree shaker to see what CSS is used and what isn't.

2) It's worth it to always minify CSS that's getting injected into JS - not minifying it doesn't actually make anything more readable (since it's always shoved into a JS string), it just makes it take up more space.

3) In spite of literally weeks trying, I've never gotten sourcemaps to work properly with CSS that's injected into JS by the postcss plugin. Since they don't help with anything anyway, it's worth it to just not generate them, as the inline comment is quite large and can't be minified/tree-shaken.

4) In a library, there really isn't a point to doing minification. Consumers should be expected to do such optimizations themselves. This makes the library a lot simpler by getting rid of the need to have dev/prod builds and switches between them (which don't typically work without help in bundlers - looking at you, `process.env` 🤨 )

So, some concrete proposals:

1) Set `preserveModules: true` for the source Rollup config (which requires pointing the output at a directory)
2) Always minify injected styles and disable source maps in the postcss plugin.
3) Remove development versions of the code bundles and produce a single artifact.

Implementing these changes, the same build as above results in these sizes, which only includes code related to the components which are actually imported:

```
8.4K headless-styles-esbuild.js
8.1K headless-styles-rollup.js
8.1K headless-styles-webpack.js
```

### Your environment 🌎

_No response_

ThatJoeMoore:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

All headless-styles components end up in one file (well, technically 4 - dev and prod builds for both ESM and CJS), which renders them un-tree-shakeable.

## What is the new behavior?

Generate separate files for each component, so that common tree-shaking algorithms can remove unused component CSS.

This also removes the 'development' build, as the only real difference between prod and dev is minification, which is more properly handled by the consuming application. It also means we no longer need bundler-specific tricks to make 'process.env' work properly.

## Other information

Fixes #940 . See that issue for more details.

### DIFF
--- a/packages/headless-styles/npm/browser.js
+++ b/packages/headless-styles/npm/browser.js
@@ -1,7 +1,0 @@
-'use strict'
-
-if (process.env.NODE_ENV === 'production') {
-  module.exports = require('./browser/index.production.min.js')
-} else {
-  module.exports = require('./browser/index.development.js')
-}
--- a/packages/headless-styles/npm/index.js
+++ b/packages/headless-styles/npm/index.js
@@ -1,7 +1,0 @@
-'use strict'
-
-if (process.env.NODE_ENV === 'production') {
-  module.exports = require('./node/index.production.min.js')
-} else {
-  module.exports = require('./node/index.development.js')
-}
--- a/packages/headless-styles/package.json
+++ b/packages/headless-styles/package.json
@@ -7,8 +7,8 @@
   "types": "npm/types/index.d.ts",
   "exports": {
     ".": {
-      "import": "./npm/browser.js",
-      "require": "./npm/index.js"
+      "import": "./npm/browser/index.js",
+      "require": "./npm/node/index.js"
     },
     "./types": {
       "types": "./npm/types/types.d.ts"
--- a/packages/headless-styles/rollup.config.mjs
+++ b/packages/headless-styles/rollup.config.mjs
@@ -40,9 +40,14 @@ function getOutputFile(isProduction, formatType, name = 'index') {
   return `npm/${folder}/${name}.${fileName}.js`
 }
 
-// rollup options
+function getOutputDir(formatType) {
+  const folder = formats[formatType].outputDir
 
-function getPlugins(isProduction) {
+  return `npm/${folder}`
+}
+
+// rollup options
+function getPlugins() {
   return [
     nodeResolve({
       extensions,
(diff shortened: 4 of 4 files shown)


caseybaggz (approved):
Great find and clean execution!

PR #958, opened 2023-01-09
Feat/879 Create useTheme Hook

### LINKED ISSUE #879: [Feature?]: add useTheme hook
### Duplicates

### Latest version

### Summary 💡

There is currently no helper for teams that are responsible for theming. We should provide a custom hook that sources and caches the theme from localStorage for teams to use.

#### Overall architecture

package: react-utils

Store and source theme from localStorage with `ps-theme` key.

- [ ] Should accept a parameter for initial value
- [ ] Should retrieve initial localStorage value (outside of hook)
- [ ] Should use localStorage value for initial value if no param given
- [ ] Should store updated value in localStorage
- [ ] Should include unit tests via component usage strategy (see other hook tests)

### Examples 🌈

Something to the effect of [this example](https://github.com/pluralsight/tva/blob/main/packages/headless-styles/sandbox/src/App.jsx#L9-L26), or whatever may be better performance wise.

### Motivation 🔦

_No response_

nicholasmackey:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

Closes #879 

## What is the new behavior?

Adds a new useTheme hook to provide a way to manage custom themes.

## Other information

[CodeSandbox Example](https://codesandbox.io/s/pando-usetheme-nywe4r?file=/src/hooks/useTheme.ts)


caseybaggz:
@nicholasmackey noticing this PR is getting a bit stale and wanted to check in. Did you mean to still have this in Draft mode? 

100% OK if you did. :)

nicholasmackey:
Hey @caseybaggz - I did intend on opening this up as a draft for my own review. Upon looking at, it I noticed that I did not include one of the requirements (Should accept a parameter for initial value.) 

If needed, I can delete this draft and reopen it once that requirement is met. I should be able to work on this more this afternoon/evening. ✌️

caseybaggz:
> Hey @caseybaggz - I did intend on opening this up as a draft for my own review. Upon looking at, it I noticed that I did not include one of the requirements (Should accept a parameter for initial value.)
> 
> If needed, I can delete this draft and reopen it once that requirement is met. I should be able to work on this more this afternoon/evening. ✌️

Ah, no worries. Work how you need. 100% was just checking in just in case it was accidentally left in the draft state. 😉

packages/react-utils/src/hooks/useTheme.ts
```diff
@@ -0,0 +1,48 @@
+import { useEffect, useState, useCallback } from 'react'
+import { isThemeValid } from '../hooks/menu/utils'
+
+const cachedTheme = localStorage.getItem('ps-theme')
+
+export function useTheme(preferredTheme?: 'light' | 'dark') {
+  const themes = {
+    light: 'light',
+    dark: 'dark',
+  }
```
nicholasmackey, line 10: I'm not certain if this is how I should handle creating a "theme map." 
I'm open to ideas here.

packages/react-utils/src/hooks/useTheme.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  const getInitialTheme = () => {
+    if (
+      preferredTheme &&
+      isThemeValid(preferredTheme) &&
+      cachedTheme === null
+    ) {
+      return themes[preferredTheme]
+    }
+    return cachedTheme || themes.light
```
nicholasmackey, line 20: If all else fails, add "light" as a fallback.

packages/react-utils/src/hooks/useTheme.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  const initialTheme = getInitialTheme()
```
nicholasmackey, line 23: Creating an intermediate variable may be unnecessary. 🤷 
I'm open to changing this.

packages/react-utils/src/hooks/menu/utils.ts
```diff
@@ -36,3 +36,11 @@ export function elementInMenu(
...
   )
 }
+
+export const isThemeValid = (theme: string) => {
+  return !!(
+    (theme === 'dark' || theme === 'light') &&
+    theme !== null &&
+    theme !== undefined
+  )
+}
```
nicholasmackey, line 46: Unsure if this formatting is up to standards. Please advise if not. 🙏

nicholasmackey:
> > Hey @caseybaggz - I did intend on opening this up as a draft for my own review. Upon looking at, it I noticed that I did not include one of the requirements (Should accept a parameter for initial value.)
> > If needed, I can delete this draft and reopen it once that requirement is met. I should be able to work on this more this afternoon/evening. ✌️
> 
> Ah, no worries. Work how you need. 100% was just checking in just in case it was accidentally left in the draft state. 😉

Thanks for your patience! This is now ready for review.

packages/react-utils/src/hooks/menu/utils.test.ts
```diff
@@ -0,0 +1,25 @@
...
+describe('isThemeValid', () => {
+  it('should return true for dark theme', () => {
+    expect(isThemeValid('dark')).toBe(true)
+  })
+
+  it('should return true for light theme', () => {
+    expect(isThemeValid('light')).toBe(true)
+  })
+
+  it('should return false if the theme is an invalid string', () => {
```
caseybaggz, line 12: I would say these last three are overkill since we are using TS, there is no need to have tests for them. Additionally, we have a strict rule to not ship `ts-expect-errors`. 😉

packages/react-utils/src/hooks/useTheme.ts
nicholasmackey: Alternatively, I thought about possibly handling it this way:

```typescript
  enum Themes {
    Light = 'light',
    Dark = 'dark',
  }

export function useTheme(preferredTheme?: 'light' | 'dark') {
  const themeMap = {
    [Themes.Light]: 'light',
    [Themes.Dark]: 'dark',
  }

  const getInitialTheme = () => {
    if (
      preferredTheme &&
      cachedTheme === null &&
      isThemeValid(preferredTheme)
    ) {
      return themeMap[preferredTheme]
    }
    return cachedTheme || themeMap[Themes.Light]
  }
  
```

Upon review, I decided that it may appear too verbose and difficult to read.

packages/react-utils/src/hooks/menu/utils.ts
```diff
@@ -36,3 +36,11 @@ export function elementInMenu(
     element && (menuElement?.contains(element) || element === triggerElement)
   )
 }
+
+export const isThemeValid = (theme: string) => {
```
caseybaggz, line 40: Since we are using TS, we really don't even need this helper. I would just kill it. Additionally, for future ref you have this in the `menu` folder, but instead, something like this should be in either a general `utils` or maybe a `useTheme` folder.

packages/react-utils/src/hooks/menu/utils.ts
caseybaggz: We use prettier which auto-formats, so there is no need to worry about that. 😸

packages/react-utils/src/hooks/useTheme.ts
```diff
@@ -0,0 +1,48 @@
+import { useEffect, useState, useCallback } from 'react'
+import { isThemeValid } from '../hooks/menu/utils'
+
+const cachedTheme = localStorage.getItem('ps-theme')
+
+export function useTheme(preferredTheme?: 'light' | 'dark') {
```
caseybaggz, line 6: Let's move the params types out of the function since it's not scalable in case more themes are added. We can use a general `type Theme` declaration which would give the benefit of exporting for the API if a user needs it for some reason.

packages/react-utils/src/hooks/useTheme.ts
caseybaggz: I'm honestly not sure we really need a map yet, but am not opposed to having one - I would just move it out of the hook so we can export it and keep the hook small (performant).

packages/react-utils/src/hooks/useTheme.ts
```diff
@@ -0,0 +1,48 @@
...
+      return themes[preferredTheme]
+    }
+    return cachedTheme || themes.light
+  }
+
+  const initialTheme = getInitialTheme()
+  const [theme, setTheme] = useState(initialTheme)
+  const { documentElement } = document
+
+  if (documentElement.getAttribute('ps-theme') === null && theme !== null) {
```
caseybaggz, line 27: This should be inside an Effect, but we honestly don't need it. By nature of how hooks/React works, this is just duplicating the work and being overwritten by the Effect below.

packages/react-utils/src/hooks/useTheme.ts
```diff
@@ -0,0 +1,48 @@
...
+
+const cachedTheme = localStorage.getItem('ps-theme')
+
+export function useTheme(preferredTheme?: 'light' | 'dark') {
+  const themes = {
+    light: 'light',
+    dark: 'dark',
+  }
+
+  const getInitialTheme = () => {
```
caseybaggz, line 12: In general, it's a good practice to keep hooks (and components) as small as possible with as less logic as possible to protect against unnecessary re-renders/commits. This function should definitely be outside of the hook and invoked immediately so the result is ready when the hook runs.

I also think it's a little overkill since we are using TS. In JS, you have to do checks like this, in TS, it's unnecessary due to the language being typed. 

Ultimately, meaning, this function should literally just get the localStorage value only. By nature, if there is no theme - our library will display `dark` as a fallback so we don't need most of this logic. 😉

packages/react-utils/src/hooks/useTheme.ts
```diff
@@ -0,0 +1,48 @@
...
+  const initialTheme = getInitialTheme()
+  const [theme, setTheme] = useState(initialTheme)
+  const { documentElement } = document
+
+  if (documentElement.getAttribute('ps-theme') === null && theme !== null) {
+    documentElement.setAttribute('ps-theme', theme)
+  }
+
+  useEffect(() => {
+    if (isThemeValid(theme)) {
```
caseybaggz, line 32: Kill the check, but also, make sure you are referencing the docs on how [design-tokens work](https://design.pluralsight.com/docs/next/development/tokens/colors#themes). We don't use the localStorage key for the html tag.

packages/react-utils/src/hooks/useTheme.ts
```diff
@@ -0,0 +1,48 @@
...
+      document.documentElement.setAttribute('ps-theme', theme)
```
caseybaggz, line 33:
```suggestion
document.documentElement.setAttribute('data-theme', theme)
```

packages/react-utils/src/hooks/useTheme.ts
```diff
@@ -0,0 +1,48 @@
...
+  const toggleTheme = useCallback(() => {
+    setTheme((prev: string) => {
+      if (prev === 'dark') {
+        return 'light'
+      }
+      return 'dark'
+    })
+  }, [])
+
+  return { theme, toggleTheme }
```
caseybaggz, line 47: Final thing, is very important regarding performance inside hooks. Since you are returning an Object that specifically uses a normal State, we need to return an memoized Object via `useMemo` which will ensure the hook doesn't force unnecessary commits when the theme state changes.

packages/react-utils/tests/hooks/useTheme.test.tsx
```diff
@@ -0,0 +1,112 @@
...
+        <div>
+          <div data-testid="theme">{theme}</div>
+        </div>
+      )
+    }
+    render(<TestComponent />)
+    expect(screen.getByTestId('theme')).toHaveTextContent('light')
+  })
+
+  test('should be light theme if the useTheme hook is provided an inital value that is invalid', () => {
```
caseybaggz, line 99: Can kill this one.

caseybaggz (changes requested):
Good initial run! I would say ultimately, definitely overkill due to bringing in some JS practices into TS (remember, they are 2 different languages) and not fully knowing our library ecosystem (which is totally fine for a new contributor).

Looking forward to the updates.

packages/react-utils/src/hooks/useTheme.ts
nicholasmackey: Noted. ✌️

packages/react-utils/src/hooks/useTheme.ts
nicholasmackey: Sure thing. I'll review that.

nicholasmackey:
> Good initial run! I would say ultimately, definitely overkill due to bringing in some JS practices into TS (remember, they are 2 different languages) and not fully knowing our library ecosystem (which is totally fine for a new contributor).
> 
> Looking forward to the updates.

Thanks for the in-depth explanations, @caseybaggz. I'll continue to work on working on your suggestions this afternoon/evening. ✌️

caseybaggz:
Just an update this week is the final week before we cut the RC release in case you want to get this feature in. No rush or pressure. We are grateful for your contribution in whatever time frame you can make it. 😉

packages/react-utils/src/hooks/useTheme.ts
nicholasmackey: Understood. I'll change that.

packages/react-utils/src/hooks/useTheme.ts
nicholasmackey: So, we'd like it to be something along the following?

`export type Themes = 'dark' | 'light'`

packages/react-utils/src/hooks/useTheme.ts
nicholasmackey: So, there may be some confusion on my part. 

Should I completely remove the line that sets the attribute on the `document` or update the hook with [your comment](https://github.com/pluralsight/pando/pull/958/files/daab804200f54118fcc46098e23e87fef7901bcf#r1068259350) below?

caseybaggz: Good question. Kill the condition (we don't need `isThemeValid`) and then, use the code below.

packages/react-utils/src/hooks/useTheme.ts
```diff
@@ -0,0 +1,28 @@
+import { useEffect, useState, useCallback } from 'react'
+
+const THEME_KEY = 'data-theme'
+const cachedTheme = localStorage.getItem(THEME_KEY)
+
+type Themes = 'light' | 'dark'
+
+export function useTheme(preferredTheme?: Themes) {
+  const [theme, setTheme] = useState(cachedTheme ?? preferredTheme)
```
caseybaggz, line 9: Since this is a condition, it would be better for us to set it using the callback version which is a little more performant in this scenario.

```suggestion
  const [theme, setTheme] = useState(() => cachedTheme ?? preferredTheme)
```

packages/react-utils/src/hooks/useTheme.ts
```diff
@@ -0,0 +1,28 @@
...
+  const toggleTheme = useCallback(() => {
+    setTheme((prev) => {
+      if (prev === 'dark') {
+        return 'light'
+      }
+      return 'dark'
+    })
+  }, [])
+
+  return { theme, toggleTheme }
```
caseybaggz, line 27: All that's left here is to use `useMemo` to return the Object since it contains a state and callback to prevent unnecessary renders/commits.

[See example](https://github.com/pluralsight/pando/blob/main/packages/react-utils/src/hooks/useAutoFormatDate.ts#L38-L45)

packages/react-utils/tests/hooks/useTheme.test.tsx
```diff
@@ -0,0 +1,83 @@
+import { render, screen, userEvent } from 'test-utils'
+import { useTheme } from '../../src'
+
+describe('useTheme', () => {
+  // Todo: getting act warnings
```
caseybaggz, line 5: We don't like to ship `Todo` comments, so I would prefer we either get those in, or just kill the comment. We actually need to do this in the entire RU lib, so it's not a big deal to just kill the comment. Also a good next issue to work on if you want. 😉

caseybaggz (changes requested):
Just a couple fine tuning things and we are good to start testing in the sandbox link! 

If you want to try your hand at that, just click the sandbox link presented by CodeSandbox in the PR and try to use your API in a real world scenario. Once you have proven the API works in the sandbox, click the "share" button and post the link as a comment on the PR for validation and historical purposes. 😸

nicholasmackey:
> Just a couple fine tuning things and we are good to start testing in the sandbox link!
> 
> If you want to try your hand at that, just click the sandbox link presented by CodeSandbox in the PR and try to use your API in a real world scenario. Once you have proven the API works in the sandbox, click the "share" button and post the link as a comment on the PR for validation and historical purposes. 😸

Sure thing. Give [this ](https://codesandbox.io/s/pando-usetheme-nywe4r?file=/src/hooks/useTheme.ts)a shot ✌️

nicholasmackey:
Here's the [CodeSandbox Example](https://codesandbox.io/s/pando-usetheme-nywe4r?file=/src/hooks/useTheme.ts) for the `useTheme` hook. 🤙

packages/react-utils/tests/hooks/useTheme.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+    render(<TestComponent />)
+    const button = screen.getByRole('button', { name: /toggle theme/i })
+
+    return { user, button }
+  }
+
+  test('should set the theme to dark', async () => {
+    const { user, button } = setup()
+    await user.click(button)
```
caseybaggz, line 31: I think if we wrap all the click actions in `waitFor` it will resolve the act warnings.

Example:
```javascript
await waitFor(() => user.click(button))
```

caseybaggz (commented):
@nicholasmackey one last tiny test update to see if it resolves the act warnings. See #996 for more information.

packages/react-utils/tests/hooks/useTheme.test.tsx
nicholasmackey: Righteous. This appears to solve the warning. 🙏

PR #976, opened 2023-01-11
refactor(hs): use inherit background for Input & Textarea

### LINKED ISSUE #957: [Feature?]: Remove input, select, textarea background-color
### Duplicates

### Latest version

### Summary 💡

Input, select, and textarea elements should no longer have a background color, but would rather have a transparent background.

### Examples 🌈

Current docs example:
CleanShot 2023-01-06 at 12 01 45@2x

Desired as seen in Figma:
Regression testing

### Motivation 🔦

Form containers will likely get a standard background to differentiate them. This change allows form elements to work on a variety of backgrounds without strange combinations of surface colors compounding and over-saturating experiences.

caseybaggz:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

closes #957 

## What is the new behavior?
Updates Input & Textarea to use `inherit` for background

## Other information
Updates Textarea to utilize base input styles.

Screenshot 2023-01-11 at 1 02 41 PM

Screenshot 2023-01-11 at 12 59 05 PM


Rykus0 (approved):

bradryanbice:
@caseybaggz Quick question on the use of `inherit`: Are we opening ourselves up to potential bugs here, in that `inherit` will not just let the parent `background` through but will actually fully inherit the value itself? For example, if an input is placed on a background with a linear-gradient applied, I'm assuming the linear-gradient value will be applied to the input box as well.

An edge case, perhaps, but also a potential problem area if inputs are placed in the right/wrong area.

bradryanbice:
@cas

> > @caseybaggz Quick question on the use of `inherit`: Are we opening ourselves up to potential bugs here, in that `inherit` will not just let the parent `background` through but will actually fully inherit the value itself? For example, if an input is placed on a background with a linear-gradient applied, I'm assuming the linear-gradient value will be applied to the input box as well.
> > An edge case, perhaps, but also a potential problem area if inputs are placed in the right/wrong area.
> 
> I mean, you could say the same about using transparent, you know what I mean? Ultimately though, this is where HS shines in being flexible for someone to correct that problem if it occurs. 👍

Yes, however I think `transparent` is more, _transparent_ (if you will) in how it will work, rather than `inherit`, if you know what I mean. Thus leading to less guess work or surprises.

bradryanbice (approved):

Issue #993, opened 2023-01-18
[Bug?]: Windows - Select options color not readable in dark theme


Rykus0:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

Options are light foreground on a light background in the dark theme

![Screenshot_20230118_102532](https://user-images.githubusercontent.com/1745718/213213443-a8c18a0b-d625-4ca9-8e5f-8544ac53491c.png)

### Expected behavior 🤔

Options have sufficient contrast to be readable.  Resetting styles to system default should be all that's needed (issue seems to be an inherited color)

### Steps to reproduce 🕹

Steps:

1. On windows, clone Pando repo
2. Run `yarn start:sandbox`
3. View "select" and expand

### Your environment 🌎

```shell
Windows 11
Edge and Firefox browsers
```

PR #998, opened 2023-01-19
Fix/993 windows option color

Rykus0:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

contributes to #993 

## What is the new behavior?
Adds props for select `option` including the fixed styles

## Other information
Docs update to follow

### DIFF
--- a/packages/headless-styles/sandbox/src/components/Select.jsx
+++ b/packages/headless-styles/sandbox/src/components/Select.jsx
@@ -65,7 +65,11 @@ function SelectField(props) {
       </Label>
       <div {...selectProps.selectWrapper}>
         <select {...selectProps.select} onChange={onChange} value={props.value}>
-          {props.children}
+          {props.options.map((option) => (
+            <option key={option} value={option} {...selectProps.option}>
+              {option}
+            </option>
+          ))}
         </select>
         <span {...selectProps.iconWrapper}>
           <ChevronDownIcon {...getIconProps(selectProps.iconOptions)} />
@@ -83,14 +87,6 @@ function SelectField(props) {
       )}
     </div>
   )
-}
-
-function SelectOptions() {
-  return selectOptions.map((option) => (
-    <option key={option} value={option}>
-      {option}
-    </option>
-  ))
 }
 
 export default function Select({ logJS }) {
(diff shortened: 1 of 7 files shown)


packages/headless-styles/sandbox/src/components/Select.jsx
```diff
@@ -65,7 +65,11 @@ function SelectField(props) {
       </Label>
       <div {...selectProps.selectWrapper}>
         <select {...selectProps.select} onChange={onChange} value={props.value}>
-          {props.children}
+          {props.options.map((option) => (
```
caseybaggz, line 68: I'm not understanding the reason you killed a component and forced the parent to inherit it's responsibility? Why not just just `SelectOptions` and update that component? This may seem like it doesn't matter, but *sometimes* in React - although rare - doing things like this does create hidden bugs when people try to use the pattern we originally had.

caseybaggz (commented):
I want to be careful in approaching fixes like we are doing here of just passively adding more stuff to established API's. Especially since this is ultimately forcing a style change just to accommodate a segment of OS users (not all of them).

I need to think on this a little more instead of us rushing out a design change to this API. In the mean time, please provide a screen shot for each browser on Mac and the windows one using Saucelabs (available in Okta).

packages/headless-styles/sandbox/src/components/Select.jsx
Rykus0: Ah, good point.  I didn't think the options component was doing much for us, and putting it here made it easier to pass the props returned from our helper.  I thought it also lended itself to a more realistic scenario.  But I can bring back the options component and refactor it to maintain that better separation.

Rykus0:
> I want to be careful in approaching fixes like we are doing here of just passively adding more stuff to established API's. Especially since this is ultimately forcing a style change just to accommodate a segment of OS users (not all of them).

Same here, which is why I brought it up in Slack first.  The other option is adding a background to the select, but that comes with its own challenges that are probably bigger.  Of the two, I thought this best since it only adds to, and does not modify, the existing API.  I also didn't want to put this off, since the options aren't readable on Windows.  But I'm totally open to adjust or hold - just let me know.

Rykus0:
Mac screenshot (Firefox)
image

Rykus0:
Windows screenshot (Firefox)
image

caseybaggz (approved):

PR #1019, opened 2023-01-20
Feat/1011 select option api

### LINKED ISSUE #993: [Bug?]: Windows - Select options color not readable in dark theme

### LINKED ISSUE #1011: [Feature?]: Create Select Option API
### Duplicates

### Latest version

### Summary 💡

Create a helper API for select `option` element props named `getSelectOptionProps`

Should return `value` and `className` props, taking `value` as an input.

Option styles need to include resetting `color` and `backgroundColor` to system defaults (see #993).

- [ ] CSS API Tests
- [ ] CSS API
- [ ] JS API Tests
- [ ] JS API

### Examples 🌈

```jsx
function Option(props) {
  const optionProps = getSelectOptionProps({
    value: props.value
  })

  return (
    <option {...optionProps}>
      {props.label}
    </option>
  )
}
```

### Motivation 🔦

Similar to Menu and MenuItem, the option is separate from the select.  Having its own API keeps a cleaner separation and improves our maintainability and scalability.

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

closes #1011 
closes #993

## What is the new behavior?
Adds API for `SelectOption` and removes option props from the `Select` API


packages/headless-styles/sandbox/src/components/Select.jsx
```diff
@@ -86,15 +87,17 @@ function SelectField(props) {
...
-    <option key={option} value={option} {...optionProps}>
-      {option}
-    </option>
+  return props.options.map((option) => (
+    <SelectOption key={option} label={option} value={option} />
   ))
 }
 
+function SelectOption(props) {
+  const selectOptionProps = getSelectOptionProps({ value: props.value })
```
caseybaggz, line 96: Yeah, this feels a lot better.

caseybaggz (commented):
I think we can either just move all this stuff back into Select and just include the new exports since it should use the shared helpers and scale in the same way with the other API.

packages/headless-styles/sandbox/src/components/Select.jsx
Rykus0: Agreed!

packages/headless-styles/src/components/Select/selectOptionCSS.module.css
```diff
@@ -0,0 +1,4 @@
+.selectOption {
```
caseybaggz, line 1: We need to just add this to the original CSS file so we don't have to add a new import for a 4 line file. 😉

packages/headless-styles/src/components/Select/selectOptionCSS.ts
```diff
@@ -0,0 +1,19 @@
+import { createClassNameProp } from '../../utils/helpers'
+import {
+  createSelectOptionProps,
+  getDefaultSelectOptionOptions,
+} from './shared'
+import type { SelectOptionOptions } from './types'
+import styles from './selectOptionCSS.module.css'
+
+const SELECT_OPTION = 'ps-select-option'
```
caseybaggz, line 9: Sorry for the confusion, I meant to just add the helper to the same selectCSS file since you don't need everything else because you can literally just re-use the select stuff down to the types.

Rykus0:
image
image
image

image
image
image

packages/headless-styles/src/components/Select/selectOptionCSS.ts
caseybaggz: Please see gridCSS as mentioned before.

caseybaggz (approved):

PR #1039, opened 2023-01-25
Fix/1038 parentless pseudo class in js

### LINKED ISSUE #1038: [Bug?]: CSS Module generation creates empty property when using psuedo-classes
### Duplicates

### Latest version

### Current behavior 😯

If we use any type of psuedo-class that's not directly tied to another class (i.e. `:is`), the code generation just defaults to an empty property which makes it buggy to use as an Object.

### Expected behavior 🤔

Should probably have a list of specific psuedo-classes that we just use it's name for the property generation. For example, `:is` would use `'is'` as the property name.

### Steps to reproduce 🕹

Steps:

1. Open up HS sandbox
2. Add a `:is()` selector to the buttonCSS.module.css file
3. Check out the generated result

### Context 🔦

Headless Styles, sandbox, Vite plugin

### Your environment 🌎

_No response_

Rykus0:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

closes #1038 

## What is the new behavior?
Pseudo-classes at the root level are no longer nested under an empty property in the resulting JS object.
Prefixed colons are removed from the property in the JS object

#### Before

```javascript
{
  '': {
    '&:root': {
      '--ps-backdrop': 'rgba(0 0 0 / 65%)',
    }
  }
}
```

#### After

```javascript
{
  'root': {
    '--ps-backdrop': 'rgba(0 0 0 / 65%)',
  }
}
```

### DIFF
--- a/packages/headless-styles/sandbox/plugins/css-to-ts/convert-rules/standard.js
+++ b/packages/headless-styles/sandbox/plugins/css-to-ts/convert-rules/standard.js
@@ -42,9 +42,11 @@ const standard = (rule, result) => {
       if (name) {
         retObj = addProperty(result, name, pseudoObj)
       } else {
+        const pseudoSelectorWithoutColon = pseudoSelector.replace(/^:+/, '')
+
         retObj = addProperty(
           result,
-          `${pseudoSelector}${secondarySelector}`,
+          `${pseudoSelectorWithoutColon}${secondarySelector}`,
           obj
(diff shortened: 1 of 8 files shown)


caseybaggz (approved):
Great work!

Issue #1001, opened 2023-01-19
[Bug?]: JS styleObjects should use `dataDisabled`


caseybaggz:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

Currently, the generated JS styles are returning the CSS syntax `data-disabled` which causes this error logged when using React:

```bash
Warning: Unsupported style property &:active:not([data-disabled='true']). Did you mean &:active:not([dataDisabled='true'])?
```

This issue also relates to Discussion #741 

### Expected behavior 🤔

At the base level, the generated files need to return the correct `dataDisabled` syntax. However, this might require an [RFC to be created](https://github.com/pluralsight/pando-rfcs) for potentially designing a refactor that ultimately solves the issue via the Discussion topic mentioned above.

### Steps to reproduce 🕹

Steps:

1. Open a sandbox using the latest next build of HS
2. use the `getJSButtonProps` helper
3. LOOK AT CONSOLE LOG to view error
4.

caseybaggz:
Also getting this on:

```bash
Warning: Unsupported style property &[data-invalid="true"]:hover. Did you mean &[dataInvalid="true"]:hover?
```

josh-degraw:
@caseybaggz @Rykus0 
I just saw a PR that addressed this just got merged, and I think that it should be reverted. I had filed an [issue](https://github.com/pluralsight/pando/issues/818) related to this console output a while ago and from looking into that I realized what was happening.

The reason that the warning is getting logged by react is because it thinks you're trying to use e.g. `&[data-invalid="true"]:hover` as an untranslated CSS property.

If you try to pass this to a plain react component, it will have no idea what to do with it, hence the warning. See this [codesandbox](https://codesandbox.io/s/smoosh-cdn-c8jcgs?file=/src/App.js) with the console error shown. I believe the reason it's suggesting to use camelcased is because react needs all style properties to be, e.g. `padding-top` becomes `paddingTop` because the former isn't valid JS.

[...]

PR #1045, opened 2023-01-30
Refactor/1001 disabled input

caseybaggz:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

contributes #1001 

## What is the new behavior?
- Updates all Input-related APIs to use new shared styles & disabled selector
- Fixes bug in composes generation that removes the (now) unnecessary `extract` helper
- Adds some JS APIs to the sandbox env but not for all due to time


packages/headless-styles/src/components/Popover/generated/popoverCSS.module.ts
```diff
@@ -39,7 +37,6 @@ export default {
     lineHeight: '1.25',
     padding: '1rem 2.5rem 1rem 1rem',
     '&::after': {
-      ...extract(tooltipCSS.tooltipContentBase, '&::after'),
```
Rykus0, line 42: In this situation, we need `extract` to merge the nested properties, so I don't think we can remove it (unless we handle the merge in the JS API)

packages/headless-styles/sandbox/plugins/css-to-ts/convert-inputs/write-file.js
```diff
@@ -127,17 +127,22 @@ function getExternalComposeFileAndValue(composes) {
 
 function getExternalEntry(externalCompose, importMap) {
   const { valueName, file } = externalCompose
-  const moduleReg = /(\.?\.\/)+(.+)\/(.+)\.module\.css/
+  // const moduleReg = /(\.?\.\/)+(.+)\/(.+)\.module\.css/
```
Rykus0, line 130: We can probably remove these comments, right?

caseybaggz: Oh, totally.

Rykus0:
Still working my way through, but it looks like there's an issue with the pagination buttons and JS tooltip (positioning)

image

image

packages/headless-styles/src/components/shared/helpers/tooltipHelpers.ts
```diff
@@ -0,0 +1,84 @@
...
+  options: Required<TooltipOptions> | Required<PopoverOptions>
+): TooltipStyleKeys<StyleModule> {
+  return {
+    contentPositionClass:
+      `${options.position}PandoPositionContent` as StyleKey<StyleModule>,
+    positionClass: `${options.position}PandoPosition` as StyleKey<StyleModule>,
+  }
+}
+
+export function getTooltipPositions(position: Position) {
```
Rykus0, line 16: With `composes` working in the JS output now, we should be able to drop these helpers and reference the styles using the same classes as the CSS API.

caseybaggz: > Still working my way through, but it looks like there's an issue with the pagination buttons and JS tooltip (positioning)
> 
> image
> 
> image

Ah, shoot! I 100% made a mental note to update that and didn't! Good catch!

caseybaggz: > With `composes` working in the JS output now, we should be able to drop these helpers and reference the styles using the same classes as the CSS API.

Do you mind adding that in? As the author, you can probably knock that out faster async wise?

caseybaggz: Fixed pagination:
Screenshot 2023-02-01 at 1 30 17 PM

Rykus0: > Do you mind adding that in? As the author, you can probably knock that out faster async wise?

Yeah, I can do that.  Sorry I missed this earlier.  I'll hop on it now.

Rykus0 (approved):

PR #1070, opened 2023-02-06
Feat/1065 make usetheme compatible for ssr

### LINKED ISSUE #1065: [Feature?]: make useTheme compatible for SSR
### Duplicates

### Latest version

### Summary 💡

Currently, `useTheme` is built assuming localStorage is always available (Client), however, this is not always the case in SSR. We should update the hook to only call the `localStorage` function if it is available to prevent SSR errors.

### Examples 🌈

_No response_

### Motivation 🔦

React utils, useTheme, SSR

caseybaggz:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

closes #1065 

## What is the new behavior?
Moves localStorage helpers into hook when the code should naturally be run in the client.

### DIFF
--- a/packages/react-utils/src/hooks/useTheme.ts
+++ b/packages/react-utils/src/hooks/useTheme.ts
@@ -1,33 +1,29 @@
 import { useEffect, useState, useCallback, useMemo } from 'react'
+import { getCachedTheme, setCachedTheme } from '../helpers/themeHelpers'
+import type { Themes } from '../types'
 
-const THEME_KEY = 'data-theme'
-const cachedTheme = localStorage.getItem(THEME_KEY)
+export function useTheme(initialTheme?: Themes) {
+  const [theme, setTheme] = useState<Themes>(initialTheme ?? 'dark')
 
-type Themes = 'light' | 'dark'
-
-export function useTheme(preferredTheme?: Themes) {
-  const [theme, setTheme] = useState(() => cachedTheme ?? preferredTheme)
+  // This mounting Effect allows this hook to avoid Hydration errors
+  // for SSR apps vs. the unreliable window check. For context see:
+  // https://github.com/facebook/docusaurus/blob/main/packages/docusaurus/src/client/browserContext.tsx
+  useEffect(() => {
+    setTheme(getCachedTheme())
+  }, [])
 
   useEffect(() => {
-    if (theme) {
-      document.documentElement.setAttribute(THEME_KEY, theme)
-      localStorage.setItem(THEME_KEY, theme)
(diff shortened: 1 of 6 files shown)


caseybaggz:
Client Side: https://codesandbox.io/s/ps-react-typescript-forked-ncnq9z?file=/package.json

CS is finally starting to support SSR apps (like NextJS) in BETA mode, so I tried testing a sandbox and it kept crashing with our API helpers, but working if I just used the code the API helpers do which makes no sense - so I'm going to assume it works on SSR and that's a bug? 🤷‍♀️

packages/react-utils/src/hooks/useTheme.ts
```diff
@@ -1,27 +1,18 @@
...
   }, [theme])
 
-  const toggleTheme = useCallback(() => {
-    setTheme((prev) => {
-      if (prev === 'dark') {
-        return 'light'
-      }
-      return 'dark'
-    })
+  const toggleTheme = useCallback((theme: Themes) => {
```
Rykus0, line 14: I think we should change the name, since this isn't toggling the theme anymore

caseybaggz: Oh yeah, good insight!

packages/react-utils/src/hooks/useTheme.ts
```diff
@@ -1,27 +1,18 @@
...
+  const [theme, setTheme] = useState<Themes>(() => {
+    return initialTheme ?? getCachedTheme()
+  })
 
   useEffect(() => {
-    if (theme) {
-      document.documentElement.setAttribute(THEME_KEY, theme)
-      localStorage.setItem(THEME_KEY, theme)
-    }
+    setCachedTheme(theme)
```
Rykus0, line 11: Won't this still try to use localStorage in SSR?  Or am I missing something?

caseybaggz: Yeah, there's no context since we can't really make a SSR sandbox yet.

useTheme is the client API, SSR API is achieved via the new helpers since each framework is different.

Rykus0: Ah, okay, thanks

packages/react-utils/src/hooks/useTheme.ts
caseybaggz: Done.

Rykus0 (approved):

PR #1076, opened 2023-02-07
Remove StyleKey construct and simplify js prop generation

josh-degraw:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

The `StyleKey` construct hid some important type information and actually increased the noise in the generated types, because it would include all possible keys from the generated module, rather than only the ones needed (e.g. instead of the relevant sentiment classes for Button it would accept any of the keys on the generated `buttonCSS.module.ts` file, which was incorrect).

## What is the new behavior?
I removed that type and updated all the places that called it to prefer using `as const` and generate the rendered keys explicitly. Basically it enforces that the `get*Classes` functions act as the source of truth for class names which will cause errors in `get*Props` functions if incorrect class names are used. I actually found and fixed a few bugs by refactoring this.


caseybaggz:
@josh-degraw mind leaving a sandbox link to validate a few CSS and JS APIs? We've updated our contribution guide to require this now on all PRs. See #1052

packages/headless-styles/src/components/Tabs/generated/tabsCSS.module.ts
```diff
@@ -9,7 +9,7 @@ export default {
   },
   tabList: {
     backgroundImage:
-      'linear-gradient(\n    to top,\n    transparent 0,\n    transparent 1rem,\n    var(--ps-border-weak) 1.0625rem,\n    var(--ps-border-weak) 1.125rem,\n    transparent 1.1875rem,\n    transparent 100%\n  )',
+      'linear-gradient(\r\n    to top,\r\n    transparent 0,\r\n    transparent 1rem,\r\n    var(--ps-border-weak) 1.0625rem,\r\n    var(--ps-border-weak) 1.125rem,\r\n    transparent 1.1875rem,\r\n    transparent 100%\r\n  )',
```
Rykus0, line 12: I think we're going to see this switch depending on the developer's platform.  Might be worth updating our helper to output with template literals, instead of quoted strings.

josh-degraw: Oh snap, I actually updated this to just replace all whitespace with a single space, which would resolve this, but it's in the `satisfies` PR. It probably merits its own PR though tbh

josh-degraw:
https://codesandbox.io/s/ps-react-typescript-urfsby

caseybaggz (approved):

caseybaggz (approved):

Rykus0 (approved):

josh-degraw:
I don't seem to have write access anymore so it won't let me push my merge commit 😖

Rykus0:
> I don't seem to have write access anymore so it won't let me push my merge commit 😖

I'll work on it this morning.  Most of these are in the stuff I added anyway, and I was anticipating these conflicts.

caseybaggz:
> I don't seem to have write access anymore so it won't let me push my merge commit 😖

Le sigh, I think this is from the GH access update they mentioned in a previous tech meeting. LSS, you are going to have to start forking again because we can't add you to the rep unless you are in the pluralsight org (not ps-dev).

Rykus0 (approved):

PR #1118, opened 2023-02-23
docs/1116 docs phase 1 redesign

### LINKED ISSUE #1116: [Docs?]: Phase 1 Redesign
### Latest version

### Summary 💡

- [ ] Re-organize the docs site to use the new IA Map. Should not include new design updates.
- [ ] Update tagging to include Classic language for easier transition to new APIs

### Motivation 🔦

See how far we can push Docusaurus to do what we need for our full re-design or if we require a fully custom solution to handle our vision.

caseybaggz:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

closes #1116 

## What is the new behavior?
Updates **only the content** of the docs to accomplish the goals of being...

1. Easier to learn about Pando
2. Easier to consume
3. Combine Design and Development resources
4. Help find the right information, faster

This PR is a part of "Phase 1" which is just a content change. Phase 2 will include a design overhaul.

## Other information
Docs Preview Deployment: https://deploy-preview-1118--famous-torte-605d60.netlify.app/


website/docs/learn/get-started/installation/add-to-website.mdx
```diff
@@ -0,0 +1,211 @@
...
+
+Our long-term vision for Pando is to have a developer experience that is just as successful as its current result. Thus, we recognize we are not there yet and plan on releasing a "core" components library in the near future to help bridge the gap for teams that may not be comfortable "owning" components while we continue to evolve Headless-styles in the future.
+
+:::
+
+If you are a team that prefers using a third-party library like [Chakra UI](https://chakra-ui.com/) or any other, you can easily extend another systems theme using Pando! :tada:
+
+### Extending Themes
+
+Out of the box, Headless-styles exports all of our CSS styles as Javascript Objects to make it as easy as possible to extend 3rd party component library's.
```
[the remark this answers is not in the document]
caseybaggz: Great point!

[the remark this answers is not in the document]
ThatJoeMoore: Ooh, yeah, that's an important point for some of the use cases we're looking at. Thanks for pointing that out!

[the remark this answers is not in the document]
caseybaggz: Done

website/docs/reference/general/themes/customizing.md
```diff
@@ -0,0 +1,264 @@
...
+
+In this example, we use Emotion to update a Link to use the Pando Link token:
+
+```javascript showLineNumbers title="Example of using a Emotion with Pando."
+import styled from '@emotion/styled'
+
+const TextLink = styled.a({
+  color: 'var(--ps-action-navigation)',
+})
+```
```
josh-degraw, line 83: This section looks like it may have the wrong title, it doesn't look related to the style objects at all

website/docs/reference/general/themes/index.mdx
```diff
@@ -0,0 +1,89 @@
...
+
+With this naming convention, you can easily use tokens based on their functionality following the naming structure:
+
+Whether you are a designer or developer, just ask yourself:
+
+### What is the purpose of this color?
+
+Every color should have a purpose, which is our **sentiment** category:
+
+- **Default:** A neutral palette for non-actionable items (i.e. page content)
```
josh-degraw, line 50: Not related to the docs specifically, but I feel like it may be worth mentioning [this discussion/suggestion](https://github.com/pluralsight/pando/discussions/812) here, as you're even using `neutral` here to describe the sentiment of the palette itself here 😜.

I really feel like using `neutral` rather than `default` will result in much less confusion for both developers and designers, and I saw that @Rykus0 also commented in agreement so I know it's not just me; is there any possibility that the naming could be re-evaluated at this point or is it pretty much set in stone by now?

caseybaggz: Well, we don't necessarily need to change the theme category (since it's still a default), however, we can update all the APIs, but that would be an insane breaking change, so maybe we can link all the `neutral` options to just use the `default` styles to allow a better experience without causing a massive "change everything in the user facing API".

website/docs/reference/general/themes/customizing.md
caseybaggz: Oh, right. Good find!

website/docs/reference/general/themes/index.mdx
josh-degraw: Yeah, I see the argument for not breaking things, I don't think linking them would be too problematic (assuming I understand what you mean by linking).

But I'd also point out that this palette is not always the default (e.g. Button defaults to the `Action` palette). I think that without the context of a specific component, it just doesn't make sense to call a palette a `default` in general because different components will necessarily have different default palettes, and for good reason. Using `neutral` would eliminate the need to know about the context of a component and still clearly communicate the purpose of that sentiment.

website/docs/reference/react-hooks/usePreloadedImage.mdx
```diff
@@ -0,0 +1,91 @@
...
+      {resource && (
+        <Suspense fallback={<Fallback />}>
+          <Image alt="test image" imgData={resource.img} />
+        </Suspense>
+      )}
+      {!resource && null}
+    </div>
+  )
+}
+```
```
josh-degraw, line 73: Can this be simplified? I'm not too familiar with Suspense but I feel like if you have to conditionally render it that it may not be fully utilized. Could this be refactored like so and work correctly?

```suggestion
function Image({ alt, src, ...props}) {
  const resource = usePreloadedImg({ alt, src })

  if(!resource) return null
  const imgProps = resource.imgData.read()
  return <img {...props} {...imgProps} />
}

export default function PreloadedImg() {
  return (
    <div className="App-container">
      <Suspense fallback={<Fallback />}>
        <Image alt="test image" src="https://source.unsplash.com/random/?face&fit=facearea&facepad=2&w=256&h=256&q=80" />
      </Suspense>
    </div>
  )
}
```

website/docs/reference/react-hooks/useRovingTabIndex.mdx
```diff
@@ -0,0 +1,61 @@
...
+
+This hook does not accept any parameters.
+
+### Returns
+
+An Object containing a `tabIndex` property of a Number value along with the event handlers needed to watch for focus events.
+
+## Troubleshooting
+
+### My tabIndex's are not updating
```
josh-degraw, line 59:
```suggestion
### My tabIndexes are not updating
```
This isn't possesive so it doesn't need an apostraphe 🙃

website/docs/reference/react-hooks/useSubmenuInteraction.mdx
```diff
@@ -0,0 +1,65 @@
...
+
+Call `useSubmenuInteraction` at the top level of the Submenu component you would like to add interactive features to:
+
+```jsx showLineNumbers {2,5,11,15}
+export function Submenu(props) {
+  const submenuInteractionProps = useSubmenuInteraction()
+
+  function onTriggerBlur(e) {
+    submenuInteractionProps.trigger.onBlur(e)
+  }
```
josh-degraw, line 40: This whole function seems unnecessary. Unless the goal is to show how to compose multiple handlers, but if that's the case I think there should be something else in the function, otherwise you could omit the additional `onBlur={onTriggerBlur}`

website/docs/reference/general/themes/index.mdx
caseybaggz: For sure, so if we "linked" them 🤷‍♀️ , it would prevent massive breaking changes that change everything foundationally from the design files to the dev packages meaning:

Tokens still use the `default` category for neutral tokens (which is fine since the names would still be the same since it's the "default use case for a page" - layout related, not actions, etc.

Then, in the API's we just add a sentiment option called "neutral" which would just use the default tokens so no breaking changes. We just need to create a CSS class map that will convert:

`.neutralBtn` to instead use `.defaultBtn` as an example (linking? Mapping? IDK...).

caseybaggz: Or...just add those classes in the CSS files, but the conversion function seems better?

website/docs/reference/react-hooks/usePreloadedImage.mdx
caseybaggz: I think it should and definitely agree that is a better pattern. We'll need to validate it in a sandbox to make sure it works.

website/docs/reference/react-hooks/useSubmenuInteraction.mdx
caseybaggz: @Rykus0 can you validate the purpose of the Trigger event on the Submenu hook?

website/docs/reference/general/themes/index.mdx
caseybaggz: #1139

website/docs/reference/react-hooks/usePreloadedImage.mdx
caseybaggz: OK, I have a working example. Because of how Suspense works, you have to be very specific where you have the Promises or else it won't work as designed. 

Pushing the update now which is the current "React" way during this "interim" period of waiting for the rest of the data management features to ship in their API (which will eventually make this hook void...but that may be like next year or longer. You never know with them.

website/docs/reference/react-hooks/useFocusTrap.mdx
```diff
@@ -0,0 +1,68 @@
...
+    setupFocusTrap()
+  }, [setupFocusTrap])
+
+  return (
+    <div {...alert.backdrop}>
+      <div {...alert.focusGuard} />
+
+      <div {...alert.wrapper} ref={wrapperRef} onClick={handleBackdropClick}>
+        <section {...alert.section} ref={ref} onKeyDown={onKeyDown}>
+  )
```
josh-degraw, line 49: I just noticed this code is invalid (no closing tags)

website/docs/reference/react-hooks/useSubmenuInteraction.mdx
Rykus0: Yeah, this was originally combining multiple callbacks, but it looks like this version doesn't include the other, so we can condense this down.  I'm currently making my way through, making small adjustments for formatting and typos and stuff, so I can update this when I get there.

PR #1133, opened 2023-03-01
Fix/1117 text button transition

### LINKED ISSUE #1117: [Bug?]: Shared Button styles missing transform transition
### Duplicates

### Latest version

### Current behavior 😯

When using the "text" option of the `usage` field, the hover transition does not exist.

### Expected behavior 🤔

There should be a smooth transition for all Buttons using the `text` usage option. The shared button styles should include `scale` in the transition list. Also need to make sure all Buttons are using `scale` property, not `transform`.

### Steps to reproduce 🕹

Steps:

1. Open dev sandbox
2. Go to Button & IconButton
3. Hover over text option
4. SEE CLUNKY TRANSITION

### Context 🔦

Headless styles, Button, IconButton, shared/button.module.css, CSS

### Your environment 🌎

_No response_

Rykus0:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

closes #1117 

## What is the new behavior?
Moves common styles to common button CSS
Applies transition to all 'text' buttons


caseybaggz (changes requested):
Bad syntax which is stopping the transition from working. You can see this when you hover over each item and they do not transition.

Screenshot 2023-03-01 at 1 11 17 PM

Rykus0:
> Bad syntax which is stopping the transition from working. You can see this when you hover over each item and they do not transition.

@caseybaggz I think you've got an older version, maybe cached or something.  That should have been fixed.

caseybaggz:
> > Bad syntax which is stopping the transition from working. You can see this when you hover over each item and they do not transition.
> 
> @caseybaggz I think you've got an older version, maybe cached or something. That should have been fixed.

The demo I click on keeps showing that syntax, so you might want to validate it's working in the local sandbox because it's definitely not in the link that is provided.

Rykus0:
> > > Bad syntax which is stopping the transition from working. You can see this when you hover over each item and they do not transition.
> > 
> > 
> > @caseybaggz I think you've got an older version, maybe cached or something. That should have been fixed.
> 
> The demo I click on keeps showing that syntax, so you might want to validate it's working in the local sandbox because it's definitely not in the link that is provided.

Weird.  I verified locally and in the codesandbox (using the codesandbox-ci link, not the one in the description, since that will have the latest build).  I'll run through them all again and make sure _I'm_ not caching anything

caseybaggz:
> 

Thanks. We need some type of validation proving this works before merging. Even if you need to create a new Sandbox and just manually import the PR packages in the package.json file - something that proves it's working.

Rykus0:
@caseybaggz I updated the demo link in the description with a new one, copied [here](https://codesandbox.io/s/ps-react-typescript-forked-ygdzlv?file=/src/App.tsx), too.

If that still doesn't work, let me know and I'll try something else

caseybaggz (approved):

PR #1172, opened 2023-03-20
fix(hs): update getErrorMessageA11yProps to use const

### LINKED ISSUE #901: [Bug?]: TS Types for some returned attributes are incorrect
### Duplicates

### Latest version

### Current behavior 😯

For attributes that have a defined subset of values, we are getting TS errors when assigning the returned props directly.  
This is primarily aria-* attributes

### getErrorMessageProps
**aria-live**: Type 'string' is not assignable to type '"off" | "assertive" | "polite" | undefined'.

### getPopoverProps
**aria-haspopup**: Type 'string' is not assignable to type 'boolean | "dialog" | "menu" | "true" | "false" | "listbox" | "tree" | "grid" | undefined'.

### getSwitchProps
**aria-hidden**: Type 'string' is not assignable to type 'Booleanish | undefined'.

### Expected behavior 🤔

Spread returned props on the target element without type errors.

### Steps to reproduce 🕹

Steps:

1.
2.
3.
4.

### Context 🔦

_No response_

### Your environment 🌎

_No response_

### LINKED ISSUE #1169: [Bug?]: Form types do not match React API
### Duplicates

### Latest version

### Current behavior 😯

When creating a Checkbox Element in Typescript, there is a clash between Pando props and React props for an InputElement which creates the error:

```bash
Interface 'CheckboxProps' cannot simultaneously extend types 'CheckboxOptions' and 'InputHTMLAttributes<HTMLInputElement>'.
  Named property 'checked' of types 'CheckboxOptions' and 'InputHTMLAttributes<HTMLInputElement>' are not identical.ts(2320)
```

Likewise, the `iconOptions` being returned do not match the `IconOptions` type so it needs to be assigned.

### Expected behavior 🤔

Pando input types should match React types for better integration.

### Steps to reproduce 🕹

[Sandbox Example](https://codesandbox.io/s/ps-react-typescript-forked-vpm47c?file=/src/Checkbox.tsx)

### Context 🔦

Headless-styles, Checkbox API, Form types, Typescript

### Your environment 🌎

_No response_

caseybaggz:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

closes #901
closes #1169 

## What is the new behavior?
Updates Pando to use correct typings for React compatibility

## Other information
Current SB: https://codesandbox.io/s/ps-react-typescript-forked-lk9w60?file=/src/ErrorMessage.tsx

### DIFF
--- a/packages/headless-styles/src/components/Pagination/shared.ts
+++ b/packages/headless-styles/src/components/Pagination/shared.ts
@@ -1,3 +1,4 @@
+import { createPandoOptions } from '../shared/defaultOptions'
 import type { ButtonOptions } from '../Button/types'
 import type { IconButtonOptions } from '../IconButton/types'
 import type { SelectOptions } from '../Select/types'
@@ -20,20 +21,27 @@ export function createPaginationClasses(options: Required<PaginationOptions>) {
 }
 
 export function createPaginationProps(options: Required<PaginationOptions>) {
+  const { size } = options
+
   return {
     buttonGroup: {},
     container: {},
     text: {},
-    buttonOptions: {
+    buttonOptions: createPandoOptions<ButtonOptions>({
       sentiment: 'default',
-      size: options.size,
-    } as ButtonOptions,
-    iconButtonOptions: {
+      size,
+    }),
+    iconButtonOptions: createPandoOptions<IconButtonOptions>({
+      ariaLabel: 'change page',
       sentiment: 'default',
-      size: options.size,
-    } as IconButtonOptions,
-    selectOptions: {
-      size: options.size,
-    } as SelectOptions,
+      size,
+    }),
+    selectOptions: createPandoOptions<SelectOptions>({
+      id: 'ps-pagination-select',
+      name: 'ps-pagination-select',
+      placeholder: '',
+      size,
+      value: '',
+    }),
   }
 }
--- a/packages/headless-styles/src/components/Pagination/shared.ts
+++ b/packages/headless-styles/src/components/Pagination/shared.ts
@@ -39,7 +39,6 @@ export function createPaginationProps(options: Required<PaginationOptions>) {
     selectOptions: createPandoOptions<SelectOptions>({
       id: 'ps-pagination-select',
       name: 'ps-pagination-select',
-      placeholder: '',
       size,
       value: '',
     }),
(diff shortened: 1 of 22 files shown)


packages/headless-styles/src/components/shared/defaultOptions.ts
```diff
@@ -39,6 +39,10 @@ export function createCheckboxFieldProps(
   }
 }
 
+export function createPandoOptions<T>(options: T) {
+  return { ...options } as T
+}
```
Rykus0, line 44: What is the benefit of using the function vs the inline assertions we had?  I can see that it's working, but curious as to why (and gaining that extra TS XP!)

caseybaggz: Good question - it's just creating a better interface for reading the code in a more "TS way". I don't think it's perfect, but I do think it's better than always using `as TYPE` everywhere.

It's like a step 1 in and long term evolution that should hopefully evolve to us writing even better APIs in the future.

caseybaggz: I ultimately think there is still a TON of TS improvements we can make, but am unsure what they are if that makes sense?

packages/headless-styles/src/components/Pagination/shared.ts
```diff
@@ -17,20 +21,27 @@ export function createPaginationClasses(options: Required<PaginationOptions>) {
...
-    },
+      size,
+    }),
+    selectOptions: createPandoOptions<SelectOptions>({
+      id: 'ps-pagination-select',
+      name: 'ps-pagination-select',
+      placeholder: '',
+      size,
+      value: '',
+    }),
```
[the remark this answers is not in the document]
caseybaggz: @Rykus0 for some reason this is not being interpreted this way in TS and is now failing because it expects the `placeholder` option due to extending FieldStates. Let me see what I can do.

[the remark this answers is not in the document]
caseybaggz: Ah, OK. I'm not 100% why, but we have to use the `Omit` class. I think it's due to having to update a deeply nested interface vs. Exclude is for shallow ones (I think).

PR #1202, opened 2023-03-28
Feat/1138 call setup focus trap internally

### LINKED ISSUE #1138: [Feature?]: useFocusTrap should call setupFocusTrap internally instead of passing to consumer
### Duplicates

### Latest version

### Summary 💡

Instead of requiring the caller to call the `setupFocusTrap` function, the API can be simplified to handle that logic internally.

### Examples 🌈

Instead of the api requiring the following:

```ts
const { ref, onKeyDown, setupFocusTrap } = useFocusTrap(triggerRef)

useEffect(() => {
  setupFocusTrap()
}, [setupFocusTrap])
```

It can instead just return `ref` and `onKeyDown` if the implementation calls setupFocusTrap internally:

```diff
+useEffect(() => {
+  handleInitFocusTrap()
+}, [handleInitFocusTrap])

return {
  ref: modalRef,
-  setupFocusTrap: handleInitFocusTrap,
  onKeyDown: handleFocus,
}
```

For scroll blocking, an argument can be passed to the hook itself: 

```ts
type FocusTrapOptions = { blockScroll?: boolean }
export function useFocusTrap(
  triggerRef: RefObject<HTMLButtonElement>,
  { blockScroll = true }: FocusTrapOptions = {}
)
```

So instead of the current style:

```ts
const { ref, onKeyDown, setupFocusTrap } = useFocusTrap(triggerRef)

useEffect(() => {
  setupFocusTrap(false)
}, [setupFocusTrap])
```
You could simplify like so:

```ts
const { ref, onKeyDown } = useFocusTrap(triggerRef, { blockScroll: false })
```

Which both simplifies and makes scroll blocking behavior a little more clear. It would also allow spreading the return value of the hook directly onto the relevant element, if that was desired.

### Motivation 🔦

I could be wrong, but I don't see any reason why that function _needs_ to be called from the consumer (other than being the current way it's set up to handle scroll blocking), and any extra work the user needs to do with an API adds friction and potential confusion or misuse, and in this case I think that friction can/should be reduced.

I'd be happy to submit a PR for this as well.

### LINKED ISSUE #1200: [Bug?]: useFocusTrap breaks if you do not provide the expected content
### Duplicates

### Latest version

### Current behavior 😯

If you are using the hook within a component that doesn't use any of the elements within the list, it will break because it expects those elements to always exist and focus on them.

### Expected behavior 🤔

Hook should conditionally call the focus on the `initFocusTrap` function via `firstItem?.focus()` so focus will not be called on something that doesn't exist.

### Steps to reproduce 🕹

Steps:

1. Remove the buttons from the ConfirmDialog sandbox component
2. Sandbox will break with error

```bash
TypeError: Cannot read properties of undefined (reading 'focus')
    at useFocusTrap.ts:53:7
```

### Context 🔦

React-utils, useFocusTrap

### Your environment 🌎

_No response_

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

closes #1138 
closes #1200 

## What is the new behavior?
useFocusTrap calls setup internally.  

## Other information
Backwards compatibility is retained by
- adding options parameter and preserving triggerRef
- continuing to return a setup function which updates state used to control the "blockScroll" behavior 

image
(other frames are TanStack router)

[Modal sandbox](https://codesandbox.io/s/basic-modal-react-typescript-forked-pd4urn)
[Popover sandbox](https://codesandbox.io/s/basic-popover-react-typescript-forked-wzxwwr)
[Confirm Dialog sandbox](https://codesandbox.io/s/basic-confirm-dialog-react-typescript-forked-5zwewb)

### DIFF
--- a/packages/react-utils/src/hooks/useFocusTrap.ts
+++ b/packages/react-utils/src/hooks/useFocusTrap.ts
@@ -2,7 +2,6 @@ import {
   useCallback,
   useEffect,
   useRef,
-  useState,
   type KeyboardEvent,
   type RefObject,
 } from 'react'
@@ -16,7 +15,6 @@ export function useFocusTrap(
   options?: FocusTrapOptions
 ) {
   const defaultOptions = defaultFocusTrapOptions(options)
-  const [blockScroll, setBlockScroll] = useState(defaultOptions.blockScroll)
   const modalRef = useRef<HTMLElement>(null)
   const selectorList =
     'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])'
@@ -55,7 +53,7 @@ export function useFocusTrap(
   )
 
   const handleInitFocusTrap = useCallback(() => {
-    setBlockScrollAttr(blockScroll)
+    setBlockScrollAttr(defaultOptions.blockScroll)
 
     if (modalRef.current != null) {
       const { firstItem } = getFocusItems()
@@ -63,12 +61,7 @@ export function useFocusTrap(
         firstItem?.focus()
       }
     }
-  }, [blockScroll, getFocusItems])
-
-  const manualInitFocusTrap = useCallback((blockScroll?: boolean) => {
-    const defaultOptions = defaultFocusTrapOptions({ blockScroll })
-    setBlockScroll(defaultOptions.blockScroll)
-  }, [])
+  }, [defaultOptions.blockScroll, getFocusItems])
 
   useEffect(() => {
     const trigger = triggerRef.current
@@ -84,7 +77,6 @@ export function useFocusTrap(
 
   return {
     ref: modalRef,
-    setupFocusTrap: manualInitFocusTrap, // deprecated
     onKeyDown: handleFocus,
   }
 }
--- a/packages/react-utils/src/hooks/useFocusTrap.ts
+++ b/packages/react-utils/src/hooks/useFocusTrap.ts
@@ -14,7 +14,7 @@ export function useFocusTrap(
   triggerRef: RefObject<HTMLButtonElement>,
   options?: FocusTrapOptions
 ) {
-  const defaultOptions = defaultFocusTrapOptions(options)
+  const defaultOptions = getDefaultFocusTrapOptions(options)
   const modalRef = useRef<HTMLElement>(null)
   const selectorList =
     'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])'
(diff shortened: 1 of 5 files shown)


packages/react-utils/src/hooks/useFocusTrap.ts
```diff
@@ -44,35 +54,51 @@ export function useFocusTrap(triggerRef: RefObject<HTMLButtonElement>) {
...
+  const manualInitFocusTrap = useCallback((blockScroll?: boolean) => {
+    const defaultOptions = defaultFocusTrapOptions({ blockScroll })
+    setBlockScroll(defaultOptions.blockScroll)
+  }, [])
 
   useEffect(() => {
     const trigger = triggerRef.current
     return () => {
-      document.body.removeAttribute('data-modal-open')
+      setBlockScrollAttr(false)
```
caseybaggz, line 76: Setting state in an effect is actually an anti-pattern and will cause error logs in specific use cases. If the code is doing this it means there is more than likely another way to do this without using state in general.

caseybaggz: IE like we were doing before - so it would be better to call the new function you created than updating a state on a component that is unmounting.

packages/react-utils/src/hooks/useFocusTrap.ts
```diff
@@ -44,35 +54,51 @@ export function useFocusTrap(triggerRef: RefObject<HTMLButtonElement>) {
...
   }, [triggerRef])
 
+  useEffect(() => {
+    handleInitFocusTrap()
+  }, [handleInitFocusTrap])
+
   return {
     ref: modalRef,
-    setupFocusTrap: handleInitFocusTrap,
+    setupFocusTrap: manualInitFocusTrap, // deprecated
```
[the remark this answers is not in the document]
Rykus0: ah, ok, that greatly simplifies things!

packages/react-utils/src/hooks/useFocusTrap.ts
Rykus0: This is setting the attribute, not the state, and I removed the state that I did have (which was to support the backward-compatible setup), so this should be all set

packages/react-utils/src/hooks/useFocusTrap.ts
```diff
@@ -6,7 +6,15 @@ import {
...
-export function useFocusTrap(triggerRef: RefObject<HTMLButtonElement>) {
+export interface FocusTrapOptions {
+  blockScroll?: boolean
+}
+
+export function useFocusTrap(
+  triggerRef: RefObject<HTMLButtonElement>,
+  options?: FocusTrapOptions
+) {
+  const defaultOptions = defaultFocusTrapOptions(options)
```
caseybaggz, line 17: This should be named `getDefaultFocusTrapOptions` since it's a getter function.

caseybaggz (approved):

Issue #1220, opened 2023-04-19
[Feature?]: add ClassNames option to all simple APIs


caseybaggz:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

To make HS have an easier interface to extend styling, we need to add a `classNames` option to all simple components:

- [x] Button
- [x] Badge
- [x] Grid
- [x] Icon Button
- [x] Icon
- [x] Tag
- [x] Textarea

For the APIs that are not composable, we will need to refactor to make them composable to compliment this feature and the future React lib. 

### Motivation 🔦

Headless Styles, ClassNames

PR #1221, opened 2023-04-19
refactor(hs): update Badge API to include classNames & composability

caseybaggz:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

contributes #1220 

## What is the new behavior?
- Adds classNames option to Badge API
- Makes Badge APIs composable

## Other information
https://codesandbox.io/s/ps-react-typescript-forked-olsptt?file=/src/App.tsx

### DIFF
--- a/packages/headless-styles/src/components/Badge/badgeJS.ts
+++ b/packages/headless-styles/src/components/Badge/badgeJS.ts
@@ -1,45 +1,27 @@
+import type { OptionProps } from '../../utils/types'
 import { createJSProps } from '../../utils/helpers'
 import {
   getDefaultBadgeOptions,
   createBadgeClasses,
-  createBadgeProps,
-  canShowIcon,
+  getIconProps,
 } from './shared'
 import styles from './generated/badgeCSS.module'
-import type { BadgeOptions } from './types'
-
-function getIconStyles(options: BadgeOptions) {
-  if (canShowIcon(options.size)) {
-    return {
-      iconWrapper: {
-        ...createJSProps(styles.badgeIcon),
-      },
-    }
-  }
-
-  return {}
-}
+import type { BadgeOptions, BadgeSize } from './types'
 
 export function getJSBadgeProps(options?: BadgeOptions) {
   const defaultOptions = getDefaultBadgeOptions(options)
-  const props = createBadgeProps(defaultOptions)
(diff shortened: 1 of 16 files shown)


packages/headless-styles/src/components/Badge/badgeJS.ts
```diff
@@ -1,45 +1,27 @@
...
-    ...styles[classnames.sizeClass],
-    ...styles[classnames.usageClass],
-  }
-
-  const iconProps = getIconStyles(defaultOptions)
-
   return {
-    ...props,
-    ...iconProps,
-    badge: {
```
Rykus0, line 40: I'm assuming we're okay with these breaking changes at this point, but want to check to make sure.

packages/headless-styles/tests/badge/badgeCSS.test.ts
```diff
@@ -1,97 +1,81 @@
...
+  test('should return icon options if size is s', () => {
+    expect(getBadgeIconProps('s')).toEqual({
+      iconOptions: {
+        ariaHidden: true,
+        ariaLabel: '',
+        customSize: '0.75rem',
+      },
+      iconWrapper: {
+        className: `${baseClass}-icon badgeIcon`,
+      },
```
Rykus0, line 78: I would group these into their own 'describes' block for `getBadgeIconProps` (maybe even its own file?), but what's your preference?  I'm okay with either one, since they're pretty small. (And I didn't think of it earlier)

packages/headless-styles/src/components/Badge/badgeJS.ts
caseybaggz: @Rykus0 good question and yes, _only_ because from this point forth, we are probably the only people who will be using HS due to the new RFC.

packages/headless-styles/tests/badge/badgeCSS.test.ts
caseybaggz: My preference is that I honestly don't care that much as long as tests exist for the things we need. For what it's worth, there are a lot of open-source teams that don't even use describe anymore - just straight up only test functions.

Rykus0 (approved):

PR #1223, opened 2023-04-20
Feat/1219 update switch styles

### LINKED ISSUE #1219: [Feature?]: Updated styles for Switch
### Duplicates

### Latest version

### Summary 💡

Switch styles have been updated:

Off state:
- Border: 1px solid `action-border`
- Handle fill: `action-text-weak`
- Handle size: Large is 12px x 12px, Small is 7px x 7px
- Large handle position is 7px from the leading edge (not including border)
- Small handle position is 5px from the leading edge (not including border)
- Hover handle fill: `action-text-inverse`
- Danger handle fill: `danger-background`

On state:
- Border: 1px solid `action-background`
- Handle fill: `action-text`
- Handle size: Large is 18px x 18px, Small is 11px x 11px
- Large handle position is 4px from the trailing edge (including the border)
- Small handle position is 4px from the trailing edge (including the border)

Transition:
- Currently set at `ease-out 200ms` (though I could be convinced of more elegant values)

Note that there is some strange use of tokens here (text tokens for handle fills, mostly). I'm hoping these can be fixed when we update `action-background-weak`, perhaps. `-inverse` use goes back to theme choices. These are needed to support the white-on-dark element in both dark and light themes.

### Examples 🌈

Figma file (branch of main UI-kit file): https://www.figma.com/file/ZmH4XsZS5WnKeo28ylM5x1/branch/UJTBe7r1J86AquJRzUJRoi/Pando-UI-kit-(Web)?node-id=2%3A676&t=CFloeKzGYGdzzCrE-1

Prototype link: https://www.figma.com/proto/ZmH4XsZS5WnKeo28ylM5x1/branch/UJTBe7r1J86AquJRzUJRoi/Pando-UI-kit-(Web)?page-id=0%3A1&node-id=6062-71617&viewport=204%2C-31%2C1&scaling=min-zoom&starting-point-node-id=6062%3A71617

I will merge this into the main UI-kit file once this is resolved.

### Motivation 🔦

- Provides a more contrasting visual difference between on and off states
- Utilizes all `action-` tokens for style (except in the label)
- Introduces more micro-animations

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

closes #1219 

## What is the new behavior?
Updates styles for Switch to match new designs

## Other information
JS tests updated to cover more cases and values

Sandbox: https://codesandbox.io/s/ps-react-typescript-forked-nypl56?file=/src/Switch.tsx


packages/headless-styles/src/components/Switch/generated/switchCSS.module.ts
```diff
@@ -37,112 +37,181 @@ export default {
...
   thumb: {
-    background: 'var(--ps-action-text)',
+    background: 'var(--ps-action-text-weak)',
     borderRadius: 'inherit',
-    height: 'var(--ps-thumb-size)',
+    height: 'var(--ps-thumb-size-off)',
     transitionDuration: '200ms',
-    transitionProperty: 'transform',
-    width: 'var(--ps-thumb-size)',
+    transitionProperty: 'transform, background-color, width, height',
```
caseybaggz, line 44: Instead of transitioning and changing both the width/height property (hella expensive), we should instead change the scale of the thumb which should still look good because it should scale down _not_ up.

This should increase the both the keyframes and the animation which is looking pretty bad on the smaller switch.

packages/headless-styles/src/components/Switch/shared.ts
```diff
@@ -8,12 +8,14 @@ import styles from './generated/switchCSS.module'
 
 export const TRACK_HEIGHT = '--ps-track-height'
 export const TRACK_WIDTH = '--ps-track-width'
-export const THUMB_SIZE = '--ps-thumb-size'
+export const THUMB_SIZE_OFF = '--ps-thumb-size-off'
```
caseybaggz, line 11: Doing so should also remove the need for these.

packages/headless-styles/src/components/Switch/switchCSS.module.css
```diff
@@ -24,28 +24,32 @@
...
   border-radius: 1000px;
   box-sizing: border-box;
   cursor: pointer;
   display: inline-flex;
   flex-shrink: 0;
+  font-size: 7px;
   height: var(--ps-track-height);
   justify-content: flex-start;
-  padding: 2px;
+  padding: 2px 4px 2px var(--ps-track-pad-start);
```
caseybaggz, line 52: Shorthand breaks our CSS rules which should have given an error for this. We need to figure out why the CI didn't call this out some other time.

packages/headless-styles/src/components/Switch/switchCSS.module.css
```diff
@@ -72,33 +72,108 @@
 }
 
 .thumb[data-checked='true'] {
-  transform: translateX(calc(var(--ps-track-width) - var(--ps-track-height)));
+  background-color: var(--ps-action-text);
+  height: var(--ps-thumb-size-on);
+  transform: translateX(
```
caseybaggz, line 77: Should also remove the need for this (now) very expensive CSS calculation.

packages/headless-styles/src/components/Switch/switchJS.ts
```diff
@@ -14,17 +15,27 @@ export function getJSSwitchProps(options?: SwitchOptions) {
...
     ...styles.track,
     ...styles[`${size}Track`],
+    ...(defaultOptions.checked && styles.track_data_checked__true),
+    ...(defaultOptions.invalid && styles.track_data_invalid__true),
     height: trackHeight,
     width: trackWidth,
   }
   const thumbStyles = {
     ...styles.thumb,
+    ...(defaultOptions.checked && styles.thumb_data_checked__true),
```
caseybaggz, line 34: Simplification to CSS should also resolve this

packages/headless-styles/src/components/Switch/switchCSS.module.css
Rykus0: I think it's because we're setting all the values.  In that case it prefers the shorthand.

Here's the message if I add all of the longhand properties:

> Expected shorthand property "padding" (declaration-block-no-redundant-longhand-properties) Stylelint [declaration-block-no-redundant-longhand-properties](https://stylelint.io/user-guide/rules/list/declaration-block-no-redundant-longhand-properties)

packages/headless-styles/src/components/Switch/switchCSS.module.css
```diff
@@ -72,33 +72,108 @@
...
+
+.thumb[data-invalid='true'] {
+  background: var(--ps-danger-background);
+}
+
+.thumb[data-invalid='true'][data-checked='true'] {
+  background: var(--ps-danger-text-inverse);
+}
+
+.thumb:not(
```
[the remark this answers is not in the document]
Rykus0: The thing is, we also want to make sure we're not applying these styles when it has those attributes, so I think we need the `not`, but I'll take a closer look.  We might still be able to reduce that.  And we should be able to condense using :is and :has like you outlined.

PR #1231, opened 2023-04-24
1230 feature refactor admontion for react library

### LINKED ISSUE #1230: [Feature?]: Refactor Admontion for React library
### Duplicates

### Latest version

### Summary 💡

To prepare for the new React lib, we need to make the Admonition API more composable to match the preferred component design.

### Examples 🌈

```jsx
<Admonition sentiment="info">
   <AdmonitionTitle>Info message</AdmonitionTitle>
   <AdmonitionDescription>This is an example of an informational admonition.</AdmonitionDescription>
   <AdmonitionCloseButton onClick={handleClose} />
</Admonition>
```

### Motivation 🔦

Headless-styles, Admonition, React

caseybaggz:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

closes #1230 

## What is the new behavior?
- Updates Admonition CSS API to be composable
- Follows #1229 standards

## Other information
https://codesandbox.io/s/ps-react-typescript-forked-393ijs?file=/src/App.tsx


packages/headless-styles/src/utils/helpers.ts
```diff
@@ -82,6 +82,10 @@ export function createJSProps<T extends NestedGeneratedStyles>(styles: T) {
   }
 }
 
+export function splitClassNameProp(classNameProp?: string) {
```
Rykus0, line 85: I feel like we should be looking for a different approach, so we don't have to have our users use an additional helper.  I think we have a couple of options, if I'm understanding correctly and not oversimplifying.
- Run all of the strings coming into `createClassNameProp` through this first
- Allow that a group of classes may be in a single string when passed in to `createClassNameProp` (I don't think it hurts anything since we're just joining an array of strings together, and any incoming classNames should already be a whitespace-delimited list of classes -- though I agree it's good to have more uniform data)
- Run incoming className string props through this internally

Rykus0: (and we could accept a string or array in the places where it may come as either)

packages/headless-styles/src/components/Admonition/admonitionJS.ts
```diff
@@ -43,7 +43,6 @@ export function getJSAdmonitionProps(options?: AdmonitionOptions) {
   return {
     ...props,
```
Rykus0, line 44: Need to add the iconOptions and iconButtonOptions, since they were removed from `createAdmonitionProps` (Unless the understanding is that the JS API users call the new `getAdmonitionCloseButtonProps` and `getAdmonitionIconProps` helpers, but I don't think that matches our established conventions)

packages/headless-styles/src/utils/helpers.ts
```diff
@@ -82,6 +82,10 @@ export function createJSProps<T extends NestedGeneratedStyles>(styles: T) {
...
+  return classNameProp?.split(' ').filter(Boolean) ?? []
```
Rykus0, line 86: Not sure which way is more efficient, but I think if you split on any repeated whitespace, you won't need to filter.
`split(/\s+/)`

packages/headless-styles/tests/utils/helpers.test.ts
```diff
@@ -94,4 +95,14 @@ opacity: 1;
     expect(transformCasing('ariaLabel', 'jsx')).toEqual('ariaLabel')
     expect(transformCasing(text, 'jsx')).toEqual(text)
   })
+
+  test('splitClassNameProp returns an array of class names', () => {
+    expect(splitClassNameProp('test')).toEqual(['test'])
+    expect(splitClassNameProp('test more names')).toEqual([
```
Rykus0, line 101: Might be good to have one like this but with multiple spaces, and similarly one with dashes and/or underscores (or possibly other acceptable, non-alpha characters) to make sure they're preserved correctly.

packages/headless-styles/src/utils/helpers.ts
caseybaggz: @Rykus0 I think there is a bit more here than seems. We are essentially doing a few things here:

1. Most importantly, we are not forcing our users to require this approach - that is an inaccurate statement
2. This helper is just that - a helper utility for *if* a user wants to adopt a String based pattern
3. Allowing our API to now be more flexible by requiring an Array of Strings (specific feedback we received from CS) and now a String alternative if you want to just use a React-like approach.

This is a starting point that very importantly extends our API capabilities vs. completely changing them which only restricts the API capabilities - which also would create a pretty big refactor that also introduces risks of recursive bugs.

caseybaggz: @Rykus0 I also do not advocate or support have multi-type parameters which only creates more logic and within the original function, increases the complexity unnecessarily, and definitely increases the risks of bugs. This approach accomplishes the same thing in a safer way.

packages/headless-styles/src/components/Admonition/admonitionJS.ts
caseybaggz: They would do exactly that because our current "established conventions" now no longer matter since after we are finished preparing for React, the JS suite is more than likely going to be killed off unless something magical happens and we hear about people using it which I doubt.

packages/headless-styles/tests/admonition/admonitionJS.test.ts
```diff
@@ -1,63 +1,82 @@
...
+
   test('should allow no props to be passed in', () => {
-    expect(getJSAdmonitionProps().wrapper.cssProps).toContain(
-      'align-items: flex-start'
-    )
-    expect(getJSAdmonitionProps().wrapper.styles.alignItems).toEqual(
-      'flex-start'
-    )
+    const props = getJSAdmonitionProps()
+    expect(props.wrapper.cssProps).toContain('align-items: flex-start')
```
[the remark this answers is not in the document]
caseybaggz: I don't think that's necessary here.

packages/headless-styles/tests/utils/helpers.test.ts
caseybaggz: Not sure how a multiple spaced class name is valid, but I'll add the other ones.

packages/headless-styles/src/utils/helpers.ts
Rykus0: Ok, I see.  And also agree on avoiding multi-type parameters

packages/headless-styles/tests/utils/helpers.test.ts
Rykus0: Multiple spaces between class names are permitted by the HTML spec, which is why I was thinking of including it as a test case

> A set of space-separated tokens is a string containing zero or more words (known as tokens) separated by one or more [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace), where words consist of any string of one or more characters, none of which are [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace).

packages/headless-styles/src/components/Admonition/admonitionJS.ts
Rykus0: Ah, ok.  That clears that up, then!

packages/headless-styles/src/utils/helpers.ts
Rykus0: Yeah.  We could make the regex easier to understand by putting it in a named variable, but I don't think it buys us much.  Though, for what it's worth, I don't find `filter(Boolean)` to be super clear, but objectively probably easier to understand than a regex (Unless you're a weirdo like me :D )

Rykus0 (approved):

Issue #1242, opened 2023-04-28
[Feature?]: Transition from CSS Modules to plain CSS


Rykus0:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Transition from CSS Modules to plain CSS.  To preserve the benefit of importing and inheriting styles, use a preprocessor (e.g., SASS).

- [x] Admonition
- [x] Alert
- [x] Avatar
- [x] Badge
- [x] Button
- [x] Checkbox
- [x] CircularProgress
- [x] ConfirmDialog
- [x] ErrorMessage
- [x] FieldMessage
- [x] FormControl
- [x] FormLabel
- [x] Grid
- [x] Icon
- [x] IconButton
- [x] Input
- [x] Menu
- [x] Modal
- [x] Pagination
- [x] Popover
- [x] Progress
- [x] Radio
- [x] Select
- [x] shared
- [x] Skeleton
- [x] Switch
- [x] Table
- [x] Tabs
- [x] Tag
- [x] Textarea
- [x] TextLink
- [x] Toast
- [x] Tooltip

Success criteria 
- Components work with SSR (particularly with NextJS 13, using the app directory feature)
- VITE plugin to generate style objects continues to function correctly

PR #1271, opened 2023-05-08
Feat/1242 remove css modules

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [x] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

contributes to #1242 

## What is the new behavior?

Adds support for scss to build
Updates Vite plugin to create JS style objects from CSS to convert css modules before processing, and scss after processing.
Converts button and icon button to scss (converted together because they referenced the same generated file)


.eslintignore
```diff
@@ -1,2 +1,3 @@
 .yarn/sdks
-website
\ No newline at end of file
+website
+packages/headless-styles/src/components/*/generated
```
caseybaggz, line 3: This shouldn't be at the global level since eslint is now scoped to each package. Is there a reason you didn't add it to the HS workspace?

```suggestion
```

packages/headless-styles/package.json
```diff
@@ -33,7 +33,10 @@
   "files": [
     "npm/**/*"
   ],
-  "sideEffects": false,
+  "sideEffects": [
+    "**/*.scss",
```
caseybaggz, line 37: Shouldn't this be the transformed CSS (i.e. `.css`)? I've never seen putting scss files in sideEffect. Literally, ever?

```suggestion
    "**/*.css",
```

packages/headless-styles/package.json
```diff
@@ -33,7 +33,10 @@
...
+    "**/*.scss-*.js"
```
caseybaggz, line 38:
```suggestion
    "**/*.css-*.js"
```

packages/headless-styles/src/components/Button/buttonCSS.scss
```diff
@@ -1,67 +1,83 @@
...
-}
-
-.btnDefaultBorder {
-  composes: pandoBtnBorder from '../shared/button.module.css';
-}
-
-.btnDefaultRadius {
-  composes: pandoBtnRadius from '../shared/button.module.css';
-}
+@use '../shared/button.scss' as *;
```
caseybaggz, line 1: Using a wildcard namespace is a not a good engineering practice and should be kept to file globs IMHO. Let's update this to be more readable.

```suggestion
@use '../shared/button.scss' as shared;
```

packages/headless-styles/src/components/Button/buttonCSS.scss
```diff
@@ -1,67 +1,83 @@
...
+
+.pando_outlineButton {
+  @include pando-btn;
+  @include pando-btn-radius;
+
+  background-color: transparent;
+  color: var(--ps-text);
+}
+
+.pando_textButton {
```
caseybaggz, line 56: We need to update this file and the others you added an underscore to, to instead use a dash like the shared files for consistency since it's what we agreed on from the start.

packages/headless-styles/src/components/IconButton/iconButtonCSS.scss
```diff
@@ -0,0 +1,95 @@
+@use '../shared/button.scss' as *;
```
caseybaggz, line 1:
```suggestion
@use '../shared/button.scss' as shared;
```

packages/headless-styles/src/components/shared/_button.scss
```diff
@@ -0,0 +1,34 @@
+@use './states.scss' as *;
```
caseybaggz, line 1:
```suggestion
@use './states.scss' as states;
```

packages/headless-styles/src/components/shared/_button.scss
```diff
@@ -0,0 +1,34 @@
...
+
+@mixin pando-btn() {
```
caseybaggz, line 3: Because this is now using SCSS, we don't need the pando name scope on all the classes for mixins.

```suggestion
@mixin base-btn() {
```

packages/headless-styles/src/components/shared/_button.scss
```diff
@@ -0,0 +1,34 @@
...
+  text-decoration: none;
+  text-transform: none;
+  transition: background-color 250ms ease-in-out, color 250ms ease-in-out,
+    scale 150ms ease-in-out;
+  user-select: none;
+  vertical-align: middle;
+  white-space: nowrap;
+}
+
+@mixin pando-btn-border() {
```
caseybaggz, line 28:
```suggestion
@mixin base-btn-border() {
```

packages/headless-styles/src/components/shared/_button.scss
```diff
@@ -0,0 +1,34 @@
...
+  border: none;
+}
+
+@mixin pando-btn-radius() {
```
caseybaggz, line 32:
```suggestion
@mixin base-btn-radius() {
```

packages/headless-styles/src/components/shared/_states.scss
```diff
@@ -0,0 +1,22 @@
+@mixin pando-default-states() {
```
caseybaggz, line 1:
```suggestion
@mixin default-states() {
```

caseybaggz (commented):
I think overall it's a good start. We need to still update a lot here. After some thinking I'm fine with keeping the underscore specifically for classes and so that way we keep the dash for values, mixins, variables, etc.

.eslintignore
Rykus0: Yeah, this is actually already in the HS workspace, but when eslint runs on the staged files, it does not have that context.  So I added this to reflect the part of that configuration that was relevant at the moment until (and if) we can configure it to use those configurations in this scenario.

packages/headless-styles/package.json
Rykus0: This instructs the bundler (rollup, currently) to include imported files that are not directly used in the code.  In this case `import 'buttonCSS.scss'` would be ignored and not processed if we didn't have it.  After we move off of rollup we will probably be able to remove the scss and add css.

packages/headless-styles/package.json
Rykus0: This was similar to the other in that, in the shipped bundle we are now importing a js file based on the parsed css file.  Without this line the CSS was not being injected when our library was used (in this case, in a CRA app).  This one should also go away once we move off of rollup.

packages/headless-styles/src/components/Button/buttonCSS.scss
Rykus0: Yeah, makes sense and I almost went back and updated them earlier.  I'll apply this to all of the cases where we're importing scss modules.

packages/headless-styles/src/components/Button/buttonCSS.scss
```diff
@@ -1,67 +1,82 @@
...
-}
-
-.btnDefaultBorder {
-  composes: pandoBtnBorder from '../shared/button.module.css';
-}
-
-.btnDefaultRadius {
-  composes: pandoBtnRadius from '../shared/button.module.css';
-}
+@use '../shared/button' as shared;
```
caseybaggz, line 1: We can kill this namespace since you are using it anymore. I'm not seeing a reason to add the extra load to create a new variable scope?

```suggestion
@use '../shared/button';
```

packages/headless-styles/src/components/IconButton/iconButtonCSS.scss
```diff
@@ -0,0 +1,92 @@
+@use '../shared/button' as shared;
```
caseybaggz, line 1:
```suggestion
@use '../shared/button';
```

caseybaggz (approved):

packages/headless-styles/src/components/Button/buttonCSS.scss
Rykus0: Yeah, we can do that.  `@extend` doesn't use the namespace, so we don't need it unless we add variables/mixins later, and we can change it then.

caseybaggz (approved):

PR #1296, opened 2023-05-10
feat(headless-styles): convert circular progress to scss

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

contributes to #1242 

## What is the new behavior?
Converts circular progress to SCSS

## Other information
[Sandbox](https://codesandbox.io/s/ps-react-typescript-forked-y4nn27)


packages/headless-styles/src/components/CircularProgress/circularProgressCSS.scss
```diff
@@ -25,32 +25,42 @@
...
   position: relative;
   vertical-align: middle;
 }
 
-.box {
+%box {
   height: 3rem;
   width: 3rem;
+
+  &.pando_indeterminateCircularProgress {
```
caseybaggz, line 39: We don't nest classes only psuedo selectors as our agreement. This is also only required for the indeterminate class which originally looks like it didn't use composes, so we should just put it back in that class and just extend box.

packages/headless-styles/src/components/CircularProgress/circularProgressCSS.scss
```diff
@@ -90,12 +90,13 @@
 
 /* sizes must be last */
 
-.mSize {
-  composes: box;
+.pando_mCircularProgress {
+  @extend %box;
```
caseybaggz, line 94: If we are using extend box in the kind classes we no longer need them here.

Rykus0: This was just to do the same as was being done in the previous css, but with sass instead.  I think it's because the determinate/indeterminate classes were put onto two different elements.  But it was somewhat confusing, so I'll work on refactoring it to be more clear.

PR #1301, opened 2023-05-11
feat(headless-styles): convert icon to scss

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

contribute to #1242 

## What is the new behavior?
Converts icon helper to use scss

## Other information
[Sandbox](https://codesandbox.io/s/ps-react-typescript-forked-n6bc11)


packages/headless-styles/src/components/Icon/iconCSS.scss
```diff
@@ -0,0 +1,28 @@
+%icon-base {
```
caseybaggz, line 1: I feel like this is the perfect scenario for us to actually use a `@mixin` to specifically create a function that interpolates through a `map` and delivers the `height/width`.

packages/headless-styles/src/components/Icon/iconCSS.scss
```diff
@@ -0,0 +1,26 @@
...
+$icon-sizes: (
+  's': 1rem,
+  'm': 1.25rem,
+  'l': 1.5rem,
+);
+
+@mixin icon-base($size: '') {
+  display: inline-block;
+
+  @if map.has-key($icon-sizes, $size) {
```
caseybaggz, line 12: I think this is a little overkill on the logic because Sass will automagically eliminate any properties that are null. So, all that is needed here is literally just the `map.get` block and not the if.

packages/headless-styles/src/components/Icon/iconCSS.scss
```diff
@@ -0,0 +1,26 @@
...
+    height: map.get($icon-sizes, $size);
+    width: map.get($icon-sizes, $size);
+  }
+}
+
+@each $name, $value in $icon-sizes {
```
caseybaggz, line 18: I would also change this to be a better name fit other than...name. Something more like `$size-key` is instantly more readable and you don't need to reference anything to know it's specifically a key of the size option for th e icon-sizes map.

caseybaggz (approved):

PR #1310, opened 2023-05-12
Feat/1242 menu scss

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

contributes to #1242 

## What is the new behavior?
Converts Menu to scss

## Other information
[Sandbox](https://codesandbox.io/s/ps-react-typescript-forked-5z2p8j)


packages/headless-styles/src/components/Menu/menuCSS.scss
```diff
@@ -46,36 +51,33 @@
...
 }
 
-.menuItem {
-  composes: pandoDefaultStates from '../shared/states.module.css';
-  align-items: center;
-  display: flex;
-  gap: 0.625rem;
-  justify-content: flex-start;
-}
+.pando_menuItem,
```
caseybaggz, line 65: I would like you to start converting your syntax to start utilizing the `:is` selector which is safer regarding the CSS compilation than the traditional list like what is being used here.

packages/headless-styles/src/components/Menu/menuPositioning.scss
```diff
@@ -0,0 +1,105 @@
...
+%vertical-start {
+  top: 0;
+}
+
+%vertical-end {
+  bottom: 0;
+}
+
+.pando_topMenu {
+  @extend %top-position;
```
caseybaggz, line 48: I think we actually want mixins here since each block is not really extending any of the placeholders, right?

packages/headless-styles/src/components/Menu/menuCSS.scss
```diff
@@ -1,3 +1,5 @@
+@use '../shared/states';
+
 @keyframes fadeIn {
```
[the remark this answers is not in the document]
Rykus0: This also has a slight scale to it.  But it would be good to pull it from the keyframes file either way.  We could create this under a new name ("pop-in?") or add a "scale-in" for just the scale and compose the animations.  I'm leaning towards "pop-in" but am not married to it.

packages/headless-styles/src/components/Menu/menuPositioning.scss
Rykus0: Agreed, mixins would be better here

packages/headless-styles/src/components/Menu/menuCSS.scss
Rykus0: I don't think `:is` benefits us in this condition, and when I tried it the converted JS output was much harder to use.  I could use `.pando_menuItem:is(button, :any-link)` which gives a better converted output, but still harder to use (`pando_menuItem['&:focus:is(button, :any-link)']` vs `pando_menuItem['&:focus']` - though we could possibly update the conversion plugin instead (later, of course)

packages/headless-styles/src/components/Menu/menuCSS.scss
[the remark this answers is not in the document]
caseybaggz: We need consistency so, `popIn` is fine to me.

packages/headless-styles/src/components/Menu/menuCSS.scss
caseybaggz: I don't mind keeping it in this file, but I feel like you are seeing the "benefit" from the developer experience only, not the browser output experience (customer). What I mean is that by using `:is` instead of the traditional list allows the browser to have a safer fallback experience if there are bad selectors mentioned.

Using the traditional list syntax that you are supporting, having bad selector values will break the entire block for everything in the list, whereas using `:is` will just ignore the single broken selector but still allow the style for all other selectors that are valid.

So, I want to challenge you to instead of putting the developer experience *before* the customer experience, you instead do the reverse when you are considering syntax decisions in the future. 👍

Rykus0: Ah, ok.  I hadn't been thinking of it that way.  In this case I was focused on the developers using the style objects as customers, but that's a great point.

I'll go with `.pando_menuItem:is(button, :any-link)` because if I use `:is(.pando_menuItem, .pando_menuItem:any-link)` it makes the JS object _very_ difficult to work with.

Rykus0: Actually, I don't think we need `:any-link` in there anymore anyway, which simplifies everything.  That was to make sure the `:link` and `:visited` states were consistent, which I don't think is an issue anymore.

caseybaggz (approved):

PR #1378, opened 2023-05-17
feat(headless-styles): convert toast to scss

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

contributes to #1242 

## What is the new behavior?
Converts Toast to SCSS

## Other information
[Sandbox](https://codesandbox.io/s/ps-react-typescript-forked-tbk3ez)


packages/headless-styles/src/components/Toast/toastCSS.scss
```diff
@@ -73,26 +74,30 @@
 
 /* sentiment */
 
-.successToast {
-  composes: toastBase;
+.pando_successToast {
```
caseybaggz, line 77: We should be able to use an iterable mixin here too and just use a sentiment map. Probably even want to make that in shared so we can always iterate over sentiments when needed? 🤔

caseybaggz: Something like `--ps-${sentiment}-surface` if it's possible in sass?

Rykus0: Definitely possible, and I think something we could get some good usage out of.

packages/headless-styles/src/components/shared/_tokens.scss
```diff
@@ -0,0 +1,15 @@
+@use 'sass:list';
+
+$sentiments: 'default', 'action', 'info', 'success', 'warning', 'danger';
+
+@function sentiments-except($excludedSentiments...) {
```
caseybaggz, line 5: Reading this is pretty hard. I actually think the TS methods are pretty nice and think we should use that same design. So, I would do a couple of things here.

1. Rename this file to be `_vars` and keep the sentiment variable in here
2. Create a new file called `_utils` and move the function in there and make 2 different functions called `pick` and `exclude` so we have the choice for whichever is easier. Params should match the TS design of the `list` to choose from and the vars you want to pick/exclude

caseybaggz (approved):

PR #1393, opened 2023-05-22
feat(hs): update Grid API to be more flexbile

### LINKED ISSUE #1391: [Feature?]: update HS Grid API
### Duplicates

### Latest version

### Summary 💡

The Grid API could be a little more flexible by adding a few more options to it.

### Examples 🌈

_No response_

### Motivation 🔦

HS, Grid

caseybaggz:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

closes #1391 

## What is the new behavior?
Updates Grid APIs to be a little more flexible

## Other information
https://codesandbox.io/s/ps-react-typescript-forked-rqjqhj?file=/src/App.tsx


packages/headless-styles/src/components/Grid/shared.ts
```diff
@@ -10,45 +19,48 @@ export const gapMap = {
...
+    colSpan: options?.colSpan ?? '1 / span 12',
+    rowSpan: options?.rowSpan ?? '',
+    style: options?.style ?? {},
   }
 }
 
-export function createGridProps(options: GridOptions) {
+export function createGridProps(options: Required<GridOptions>) {
+  const { areas, cols, rows } = options
+  const gap = gapMap[options.gap as keyof typeof gapMap] ?? 0
```
Rykus0, line 43: Any reason we're not using `GridGap` instead of `keyof typeof gapMap`?  We could update the typing of `gapMap` too, and remove this assertion altogether.

packages/headless-styles/src/components/Grid/types.ts
```diff
@@ -1,16 +1,23 @@
...
+    Pick<HTMLAttributes<HTMLDivElement>, 'style'> {
+  area?: keyof GridOptions['areas'] | string
+  colSpan?: string
+  rowSpan?: string | null
 }
 
 // types
 
-export type GridGap = 6 | 8 | 12 | 16 | 32
+export type GridGap = 6 | 8 | 12 | 16 | 32 | number
```
Rykus0, line 23: Since we're constraining this to what's in the gap map, we should probably remove `| number`

packages/headless-styles/src/components/Grid/shared.ts
caseybaggz: @Rykus0 I tried that first and TS still wasn't happy

packages/headless-styles/src/components/Grid/shared.ts
```diff
@@ -10,45 +19,48 @@ export const gapMap = {
...
-    }
-  }
-
   return {
     style: {
-      gridColumn: `span ${colSpan} / span ${colSpan}`,
+      ...options.style,
+      gridArea: options.area,
+      gridColumn: colSpan,
+      gridRow: rowSpan,
```
Rykus0, line 63: The naming feels confusing to me here, mapping the span to the column/row property.  But I also understand the flexibility we're going for.   Would you want to rename to col/row, or maybe some other options to balance the ease-of-use and flexibility?

packages/headless-styles/tests/grid/gridCSS.test.ts
```diff
@@ -45,18 +50,18 @@ describe('Grid CSS', () => {
...
+        getGridItemProps({ colSpan: '3 / span 8' })?.style.gridColumn
+      ).toEqual('3 / span 8')
     })
 
     test('should update the rowSpan when option given', () => {
-      expect(getGridItemProps({ rowSpan: 4 })?.style.gridArea).toEqual(
-        'span 4 / span 12 / span 4 / span 12'
-      )
+      expect(
+        getGridItemProps({ rowSpan: '4rem auto 6rem' })?.style.gridRow
```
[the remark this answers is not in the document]
caseybaggz: That's the same thing I used in the sandbox and it works fine? 🤷‍♀️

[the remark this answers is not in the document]
Rykus0: ah, I see.  Looks like that's for the grid row template, not for the grid item row, so this is a different case.  The API will accept it, but it wouldn't be valid in the css

packages/headless-styles/src/components/Grid/types.ts
caseybaggz: I understand that perspective, but I put this in here so there is more flexibly when people need something outside of the map.

packages/headless-styles/src/components/Grid/shared.ts
caseybaggz: Valid statement

packages/headless-styles/src/components/Grid/types.ts
Rykus0: Or should it be `| string` since a custom value would need a unit also... though we don't want to mix types, so maybe also convert the other values to strings?

caseybaggz: Let me think on this some more. I bet there is a way to use both and have good typing.

packages/headless-styles/src/components/Grid/shared.ts
caseybaggz: Done in best way possible I know for now? 🤷‍♀️

packages/headless-styles/src/components/Grid/types.ts
caseybaggz: Updated to use string values and dynamic typing.

packages/headless-styles/src/components/Grid/shared.ts
caseybaggz: Done.

Rykus0 (approved):

PR #1426, opened 2023-05-26
Feat/1418 get flex props

### LINKED ISSUE #1418: [Feature?]: add getFlexProps API
### Duplicates

### Latest version

### Summary 💡

Add an API that functions similar to the existing `getGridProps` API but for flexbox

### Examples 🌈

_No response_

### Motivation 🔦

_No response_

Rykus0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

closes #1418 

## What is the new behavior?
Add flex API, similar to grid API for flex-based layouts

## Other information
[Sandbox](https://codesandbox.io/s/ps-react-typescript-forked-emfriv?file=/src/App.tsx)

### DIFF
--- a/packages/headless-styles/src/components/Flex/flexCSS.scss
+++ b/packages/headless-styles/src/components/Flex/flexCSS.scss
@@ -1,0 +1,8 @@
+.pando_flexContainer {
+  display: flex;
+  width: 100%;
+}
+
+.pando_flexItem {
+  min-width: 0;
+}
--- a/packages/headless-styles/src/components/Flex/flexCSS.scss
+++ b/packages/headless-styles/src/components/Flex/flexCSS.scss
@@ -4,5 +4,5 @@
 }
 
 .pando_flexItem {
-  min-width: 0;
+  min-width: initial;
 }
--- a/packages/headless-styles/src/components/Flex/shared.ts
+++ b/packages/headless-styles/src/components/Flex/shared.ts
@@ -1,9 +1,10 @@
 import { type CSSProperties } from 'react'
 import type {
-  FlexOptions,
-  FlexItemOptions,
-  FlexGapPreset,
+  FlexDirection,
   FlexGap,
+  FlexGapPreset,
+  FlexItemOptions,
+  FlexOptions,
 } from './types'
 
 const gapMap: Record<FlexGapPreset, string> = {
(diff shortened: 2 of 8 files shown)


packages/headless-styles/src/components/Flex/flexCSS.scss
```diff
@@ -0,0 +1,8 @@
+.pando_flexContainer {
+  display: flex;
+  width: 100%;
+}
+
+.pando_flexItem {
+  min-width: 0;
```
caseybaggz, line 7:
```suggestion
  min-width: initial;
```

packages/headless-styles/src/components/Flex/shared.ts
```diff
@@ -0,0 +1,77 @@
...
+  '12': '0.75rem',
+  '16': '1rem',
+  '32': '2rem',
+}
+
+function getGapValue(gap: FlexGap) {
+  if (Object.prototype.hasOwnProperty.call(gapMap, gap)) {
+    return gapMap[gap as FlexGapPreset]
+  }
+  return gap
```
caseybaggz, line 22: I think here we should probably (we can also do a Stacked PR into this branch for doing the same thing in Grid) instead return the REM maths, then just kill the type and `gapMap` since we aren't getting any value from the typing itself:

```suggestion
  return `${gap / 16}rem`
```

caseybaggz: Then, we can change the type to back to `number`.

packages/headless-styles/tests/flex/flexCSS.test.ts
```diff
@@ -0,0 +1,125 @@
...
+
+    test('should accept custom styles', () => {
+      expect(
+        getFlexProps({ style: { backgroundColor: 'yellow' } })?.style
+          .backgroundColor
+      ).toEqual('yellow')
+    })
+  })
+
+  describe('getFlexItemProps', () => {
```
caseybaggz, line 93: Let's kill this overly nested test and instead just do 2 root level describes using the normal `Flex` and `Flex Item`.

packages/headless-styles/src/components/Flex/shared.ts
Rykus0: Ok.  I had it this way to allow entry of any custom value while also providing the "presets." Though it looks like the intellisense doesn't show the presets the way it is now  If I change the presets to numbers, then they show up and we'd still be able to accept both... but I don't know if that causes too much confusion.

If we just go with the number, then the only options would be px converted to rems.  If that's enough, then we can do that, but I thought we wanted to have some more flexibility so I went this route.

Rykus0: Ok, thanks. I'll make the suggested updates, then stack the corresponding change to Grid

caseybaggz (approved):

PR #1806, opened 2023-10-18
1754 feature install pkgs

### LINKED ISSUE #1754: [Feature?]: pando cli setup: install dependencies
### Duplicates

### Latest version

### Summary 💡

- [x] running `pnpx pando setup` should
  - [x] detect & confirm PM pref (bun, npm, yarn, pnpm)
  - [x] install `@pluralsight/{react,icons}`
  - [x] install `@pluralsight/panda-preset`
  - [x] install `@pandacss/dev`
  - [x] install `postcss` (if not already installed)
  - [x] install `autoprefixer` (if not already installed)

### Examples 🌈

_No response_

### Motivation 🔦

_No response_

kathydlavoie-ps:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

closes #1754


tests/pando/mocks.ts
kathydlavoie-ps, line 1: I did not know if there was a way to access the `consts` file in the pando package from the root tests so I just duplicated these - I figure it might be good to have it hardcoded in tests so that tests alert us to if something changes anyways though 🤷

tests/pando/harness.ts
kathydlavoie-ps, line 1: I was having trouble deciding whether to put things in here or in `helpers.ts` so I just nixed this one

packages/pando/src/shared/types.mts
```diff
@@ -0,0 +1,8 @@
+export type CLIOperation = 'setup' | 'update' | undefined
+
+export enum PMOptions {
```
kathydlavoie-ps, line 3: I know that I had removed an enum based on prior feedback, but I like that this can function as a string and a type. I was uncertain of how kind of negotiate between having a string const but also wanting a type while reducing duplication 🤷 I just intro'd a type above (`CLIOperation`) because it is actually possible that it is undefined - otherwise I may have converted that into an enum as well.

tests/pando/harness.ts
kathydlavoie-ps, line 1: I was having trouble deciding whether to put things in helpers or this file so I just nixed this one 🤷

tests/pando/mocks.ts
kathydlavoie-ps, line 1: I did not know if there was a way to access the consts file in the pando package from the root tests so I just duplicated these - I figure it might be good to have it hardcoded in tests so that tests alert us to if something changes anyways though 🤷

caseybaggz (approved):

caseybaggz (dismissed):

packages/pando/src/shared/prompts.mts
```diff
@@ -1,9 +1,31 @@
-export function getCliError(cmd: 'setup' | 'update') {
-  return `Oh no! There was a problem running the ${cmd} command. Please report this error to the  Pando Team by submitting an issue on Github: https://github.com/pluralsight/pando/issues/new?assignees=&labels=bug%2Cneeds+triage&projects=&template=bug.yml&title=%5BBug%3F%5D%3A+.`
+import { CLIOperation, PMOptions } from './types.mts'
+
+export function getCliError(cmd?: CLIOperation) {
+  return `Oh no! There was a problem running the${
+    cmd ? ' ' + cmd + ' ' : ' '
```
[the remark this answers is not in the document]
kathydlavoie-ps: can you elaborate on "outdated syntax"?

packages/pando/src/shared/types.mts
kathydlavoie-ps: @caseybaggz I did not know that it was the current standard to avoid enum. It has been a standard in some prior work that I had done, but reading more into it now I understand more of the concerns. What I was trying to achieve with this implementation was one location to cover both a const & a type, like for example avoiding defining a type separately from consts. But I will try out a different pattern and see what you think.

packages/pando/src/shared/types.mts
kathydlavoie-ps, line 1: it feels odd to import consts to generate types in another file, but types don't fit into the filename `consts` - any suggestions on how to better combine these concerns is welcome 🙏

tests/pando/helpers.ts
```diff
@@ -13,3 +17,32 @@ export function setup(command?: 'update' | 'setup') {
...
+  spawnSync(['bun', 'install'])
+}
+function getPandoPackageJson() {
+  return relative(import.meta.dir, 'packages/pando/package.json')
+}
+export function readPackageJson() {
+  return readFileSync(getPandoPackageJson(), 'utf-8')
+}
+
+export const DOWN = '\x1B\x5B\x42'
```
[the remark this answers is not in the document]
kathydlavoie-ps: I am using these "arrow action" strings to input to the spawn process to simulate the user clicking through the cli

packages/pando/src/shared/const.mts
```diff
@@ -1,2 +1,31 @@
...
+export const NPM = 'npm'
+
+// lockfiles
+export const BUNLOCK = 'bun.lockb'
+export const PNPMLOCK = 'pnpm-lock.yaml'
+export const YARNLOCK = 'yarn.lock'
+export const NPMLOCK = 'package-lock.json'
+
+// packages
+export const pandoPkgs = [
```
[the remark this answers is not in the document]
kathydlavoie-ps: I am using the full package name to actually install the packages as well as in some messaging etc. I think it is easier to keep the full name together rather than concat `@pluralsight/` on wherever necessary - are you thinking that it would be better to exclude `@pluralsight/` from confirmation messaging? ex: 

```Installing Pando packages react, icons, panda-preset using bun add...```

vs current 

```Installing packages @pluralsight/react, @pluralsight/icons, @pluralsight/panda-preset using bun add...```

File: packages/pando/package.json [hunk 5]
```diff
@@ -30,12 +30,23 @@
...
     "@types/inquirer": "^9.0.4",
-    "inquirer": "^9.2.11",
     "tslib": "latest",
     "yargs": "^17.7.2"
   },
   "devDependencies": {},
+  "exports": {
+    ".": "./index.mts",
+    "./package.json": "./package.json",
+    "./shared/const.ts": "./src/shared/const.mts",
```
caseybaggz, line 44: This isn't really taking advantage of what this feature has the ability to do for exporting. I would recommend updating all the aliases to be a shortcut path vs. the literal path to help keep the codebase clean for the user.

This:
```javascript
import { THING } from '@pluralsight/pando/const` 
```
is much better (and is the common standard) than this:
```javascript
import { THING } from '@pluralsight/pando/shared/const.ts`
```

The job of an API is to make things easier (including a clean/readable codebase) not harder (like using a literal file path down to the extension). 😉

caseybaggz: The package.json path is the only one I would keep the same but would only include it if you actually use it.

packages/pando/src/index.mts
```diff
@@ -1,43 +1,39 @@
...
 import { hideBin } from 'yargs/helpers'
-import inquirer from 'inquirer'
-import { pandoSetup } from './pandoSetup.mts'
-import { pandoUpdate } from './pandoUpdate.mts'
-import { SETUP, UPDATE } from 'shared/const.mts'
-import { getCliError, usage, welcome } from 'shared/prompts.mts'
+import select from '@inquirer/select'
+import { pandoSetup } from './setup/pandoSetup.mts'
+import { pandoUpdate } from './update/pandoUpdate.mts'
+import { CLIOPERATION } from 'shared/const.mts'
```
caseybaggz, line 6: These should use relative paths bc they are local files. Absolute paths should be reserved for third-party modules to help the brain read the code faster.

packages/pando/src/setup/pandoSetup.mts
```diff
@@ -0,0 +1,25 @@
+import { step1 } from './step1/step1.mts'
+import { step2, step3 } from './step2And3/packageInstall.mts'
+
+export async function pandoSetup() {
+  console.log('Welcome to Pando setup')
+  let bunSpawnInstallScript: string[]
+
+  try {
```
caseybaggz, line 8: This triple `try/catch` is unnecessary, especially because nothing is happening in the catch which means...ultimately - it's useless.

I would prefer a single `try/catch` and actually throw a meaningful error in the `catch` block.

packages/pando/src/setup/pandoSetup.mts
```diff
@@ -0,0 +1,25 @@
+import { step1 } from './step1/step1.mts'
+import { step2, step3 } from './step2And3/packageInstall.mts'
+
+export async function pandoSetup() {
+  console.log('Welcome to Pando setup')
```
caseybaggz, line 5: Reverse these please and space between (I would suggest, but I can't multi-line suggest 🙈 )

packages/pando/src/setup/step1/prompts.mts
```diff
@@ -0,0 +1,7 @@
+import { PMOptions } from 'shared/types.mts'
+
+export const step1Msg = 'Step 1: Confirm your package manager'
+
+export function confirmDetectedPm(lockFileName: string, pm: PMOptions) {
+  return `We've detected a ${lockFileName} file. Would you like us to install dependencies using ${pm}?`
```
caseybaggz, line 6:
```suggestion
  return `Looks like you are using ${pm}. Would you like us to install dependencies with this package manager?`
```

packages/pando/src/setup/step1/step1.mts
```diff
@@ -0,0 +1,48 @@
...
+export async function step1() {
+  console.log(step1Msg)
+  const detectedLockfile = detectLockfile()
+  const detectedPm = detectedLockfile && detectPm(detectedLockfile)
+  const selectedPm = await confirmDetectedOrManuallySelect(
+    detectedLockfile,
+    detectedPm,
+  )
+
+  if (selectedPm) {
```
caseybaggz, line 20: This should instead use a `try/catch` and provide a meaningful error message.

packages/pando/src/setup/step1/step1.mts
```diff
@@ -0,0 +1,48 @@
...
+  detectLockfile,
+  detectPm,
+  manuallySelectInstallScript,
+} from 'shared/utils.mts'
+import { confirmDetectedPm, step1Msg } from './prompts.mts'
+import confirm from '@inquirer/confirm'
+import { Lockfiles, PMOptions } from 'shared/types.mts'
+import { getInstallScript } from 'shared/const.mts'
+
+export async function step1() {
```
caseybaggz, line 11: I think we can do better with this name. Having no context - `step1` means nothing. What is it? IDK, but it's step 1! Instead, it should be named what step one actually is responsible for - `validatePackageManager`.

packages/pando/src/setup/step1/step1.mts
```diff
@@ -0,0 +1,48 @@
...
+  )
+
+  if (selectedPm) {
+    return selectedPm
+  } else {
+    throw new Error()
+  }
+}
+
+async function confirmDetectedOrManuallySelect(
```
caseybaggz, line 27: These last two helpers read odd and I think it's because we are calling promises based on the results of parameters which doesn't make sense to me. Maybe even with the naming of the functions too? I'm honestly unsure but something is not right with the readability.

The naming of this first function should give an indicator it is doing more than just a single responsibility because it has "or" in the name. This means the code can and should be refactored which might even allow us to eliminate some of this stuff which will help with complexity.

packages/pando/src/setup/step2And3/packageInstall.mts
```diff
@@ -0,0 +1,31 @@
+import { pandoPkgs, reqdDepPkgs } from 'shared/const.mts'
+import { confirmProceed, denyProceed } from 'shared/prompts.mts'
+import { spawn } from 'bun'
+import { pkgInstallMsg, step2Msg, step3Msg } from './prompts.mts'
+
+async function confirmAndInstall(script: string[], pkgs: string[]) {
```
caseybaggz, line 6: Same feedback here. Also, since I haven't updated the eslint yet, the imports should use the following order:

1. dependencies
2. Local by relative path location (furthest > closest)

packages/pando/src/setup/step2And3/prompts.mts
```diff
@@ -0,0 +1,7 @@
+export const step2Msg = 'Step 2: Install Pando Packages'
```
caseybaggz, line 1: We don't need the `Msg` appended to the name since it's being imported from the `prompts` file which gives the context - also, note the disconnect between the file name (prompts) and the syntax being chosen (msg) - those should be consistent if you use them.

I also think the same idea applies here. For example `INSTALL_PROMPT` is more meaningful than `step2Msg`

packages/pando/src/shared/const.mts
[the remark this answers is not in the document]
caseybaggz: Yes. If people are using Pando, they know which packages are going to be associated with them from our ecosystem.

packages/pando/src/shared/prompts.mts
```diff
@@ -1,9 +1,27 @@
-export function getCliError(cmd: 'setup' | 'update') {
+import { CLIOperation } from './types.mts'
+import confirm from '@inquirer/confirm'
+
+export function getCliOperationError(cmd: CLIOperation) {
```
caseybaggz, line 4: We should extract these into a file called `errors.mts`

caseybaggz (commented):
Let's pair on this one. 👍

packages/pando/src/setup/pandoSetup.mts
kathydlavoie-ps: oh duh 🤦 thanks

caseybaggz:
Going to close for now since we have pivoted. Branch will always remain in history for reference if we need.

PR #1883, opened 2023-12-05
1880 sandbox tests

kathydlavoie-ps:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

Closes #1880 

## What is the new behavior?

- removes data-testid from sandbox test
- standardizes whitespace within tests

### DIFF
--- a/tests/sandbox/AppBar.test.tsx
+++ b/tests/sandbox/AppBar.test.tsx
@@ -10,35 +10,32 @@ describe('AppBar', () => {
 
   test('contains home link', () => {
     render(<AppBar />)
-    const homeLink = screen.getAllByRole('link')[0]
+    const homeLink = screen.getByRole('link', { name: 'Pando Dev Sandbox' })
 
     expect(homeLink).toHaveAttribute('href', '/')
-    expect(homeLink).toHaveTextContent('Pando Dev Sandbox')
   })
 
   test('contains docs link', () => {
     render(<AppBar />)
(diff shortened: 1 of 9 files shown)


tests/sandbox/AppBar.test.tsx
```diff
@@ -4,21 +4,38 @@ import AppBar from 'sandbox/app/components/AppBar'
...
+    expect(screen.getByText('Pando Dev Sandbox')).toBeInTheDocument()
   })
-  test('contains intended string', () => {
+
+  test('contains home link', () => {
     render(<AppBar />)
-    expect(screen.getByText('Pando Dev Sandbox')).toBeInTheDocument()
+    const homeLink = screen.getByRole('link', { name: 'Pando Dev Sandbox' })
+
+    expect(homeLink).toHaveAttribute('href', '/')
```
caseybaggz, line 15: Because all the links are static I'm not sure we need to test the value of the href because not having one would throw a lint error which would break the test naturally.

So, I feel like for all the links just testing this is fine for our use case:
```
expect(screen.getByRole('link', { name: 'Pando Dev Sandbox' })).toBeInTheDocument()
```

Also, keep in mind (per our style guide) we don't want to create variables unless there is a need (3 or more repititions).

caseybaggz (approved):

PR #1910, opened 2023-12-14
chore: add Brad to owner for docs v2

kathydlavoie-ps:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes
- [x] Other... Please describe: code owner update

Currently only @caseybaggz & @kathydlavoie-ps have the ability to approve pr's to pando docs as code owners.

## What is the new behavior?

This also adds @bradryanbice

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -8,3 +8,6 @@
 
 # Design doc changes
 website/docs/design/*  @bradryanbice
+
+# Pando docs content changes
+docs/src/content/docs/*  @bradryanbice


caseybaggz (commented):
I don't feel comfortable approving this unless @bradryanbice gives an approval understanding the consequences of this update since this directly impacts him. 

I don't feel that it's right that either you or I are the ones to approve this submission - it should be @bradryanbice 's choice.

bradryanbice:
I think the only distinction I would want to make, if it's even possible, would be to not be tagged or required to review every engineering-focused doc page. I'm also not entirely sure this is needed as I'm assuming as a small team (so far) any suggestions or observations I have for merged content can always be brought back up.

I appreciate the sentiment here, definitely. However I also want to ensure you're all able to move as fast as possible. I'll follow your lead here on what you think is best, knowing that I don't mind not being tagged on everything. Or maybe we can revisit if it feels like I'm being left out of major additions I'd like to be tagged on?

PR #1911, opened 2023-12-14
1887 feature button component migrate headless styles to pandacss

### LINKED ISSUE #1887: [Feature?]: Button component: migrate headless styles to pandacss
### Duplicates

### Latest version

### Summary 💡

Migrate Button component to use a Panda CSS recipe

### Examples 🌈

_No response_

### Motivation 🔦

the Pando Panda CSS Preset will replace the headless styles package in the v1 work and the headless styles package will be deprecated, so all components that use the headless styles package need to be migrated.

caseybaggz:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

closes #1887 

## What is the new behavior?
- kills dependabot checks for deprecated libs
- Adds panda-preset button recipe
- Updates React lib to use recipe
- Updates React API to be SSC compatible
- Updates Sandbox Button page to test for all accounts visually
- Updates Sandbox settings to match current Panda rec

Tests coming in next PR 👈 

## Other information
Screenshot 2023-12-14 at 10 38 56 AM


packages/react/src/components/Button.tsx
```diff
@@ -3,55 +3,44 @@
...
-    disabled,
-    size,
-    sentiment,
-    usage,
-  })
-  const iconProps = getIconProps(getButtonIconOptions(size))
-  const StartIcon = startIcon
-  const EndIcon = endIcon
-
+  const { children, palette, usage, size, startIcon, endIcon, ...nativeProps } =
```
kathydlavoie-ps, line 29: Why destructure here? Generally we destructure if a variable is used more than 2 times. Is it to avoid syntax like:

```
button({
  palette: props.palette
  usage: props.usage
  ...
})
```

& would this affect when we allow destructured vars vs not?

packages/react/src/components/Button.tsx
```diff
@@ -3,55 +3,44 @@
...
-  getIconProps,
-  splitClassNameProp,
-} from '@pluralsight/headless-styles'
-import type { ButtonOptions } from '@pluralsight/headless-styles/types'
+import type { Palettes, Sizes } from './shared/types'
+import { button } from '@/styled-system/recipes'
+
+export type ButtonPalette = Exclude<Palettes, 'success' | 'info' | 'warning'>
+export type ButtonUsage = 'filled' | 'outline' | 'text'
+export type ButtonSize = Exclude<Sizes, 'xs' | 'sm' | 'xl'>
```
kathydlavoie-ps, line 15: curious - is there a reason why the `Exclude<...` is preferred? Especially as in the `ButtonSize` usage more types are excluded than included, it may be more readable to use [Pick](https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys) `Pick<Sizes, 'md' | 'lg'>` It could improve readibility of the type rather than to analyze the original type and subtract what is not there.

packages/react/src/hooks/useTheme.ts
```diff
@@ -32,6 +35,10 @@ export function useTheme<T extends string, M extends string>(
     setMode(newMode)
   }, [])
 
+  useEffect(() => {
+    setIsClient(true)
+  }, [])
```
kathydlavoie-ps, line 40: jw - if we are only running this state change once on initial render, why is the initial assignment (on line 22) `false`?

kathydlavoie-ps (commented):
couple q's!

packages/react/src/components/Button.tsx
caseybaggz: We need to destructure here in order to prevent a React error from non nativeProps being passed to the react jsx component. This is also why we are specifically using `...nativeProps` to help ensure that is understood. 👍

packages/react/src/hooks/useTheme.ts
caseybaggz: This is a NextJS requirement and straight from their docs in order to prevent hydration errors with their building process.

packages/react/src/components/Button.tsx
caseybaggz: Pick is reserved for Objects and interfaces and cannot be used for unions.

kathydlavoie-ps (approved):
🚀 🤩

PR #2033, opened 2024-01-18
feat(docs): add danger palette & references

### LINKED ISSUE #1964: [Feature?]: (docs) danger palette reference page
### Duplicates

### Latest version

### Summary 💡

Create an danger palette reference page similar to #1926, which was for the neutral palette. Should be based off of [Figma UI Kit](https://www.figma.com/proto/uJtPfI38D9i8iQg0UGK2E0/Pando-Design-Guidelines?node-id=483-20797&starting-point-node-id=6%3A11626)

### Examples 🌈

`/reference/colors/neutral-palette` in new docs

### Motivation 🔦

Each component of the color palette will need its own page for reference.

### LINKED ISSUE #2005: [Bug?]: Typo in Explanation -> Color -> Neutral
### Duplicates

### Latest version

### Current behavior 😯

On https://pando.pluralsight.com/explanation/color/, I see "A palette for non-actionale items".

### Expected behavior 🤔

I expect to see "A palette for non-actionable items"

### Steps to reproduce 🕹

No repro steps since this issue is in the docs 😄 

### Context 🔦

_No response_

### Your environment 🌎

_No response_

kathydlavoie-ps:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] infrastructure changes

closes #1964 
closes #2005 (bug)

## What is the new behavior?

* Adds danger palette ref page & navigation
* Updates color titles in to a font color of each color scheme

## Other information

* This is the last color reference page - after this color will be complete, save for clarifying details for colors on specific sentiment reference pages

https://github.com/pluralsight/pando/assets/146388897/edc5f3d9-a32b-489f-a844-09c08cfdbb06

### DIFF
--- a/docs/src/content/docs/explanation/color.mdx
+++ b/docs/src/content/docs/explanation/color.mdx
@@ -5,43 +5,60 @@ hero:
   tagline: Usage guidelines and naming references
 ---
 
+import { textLink } from '@/styled-system/recipes'
+import { css } from '@/styled-system/css'
+
 ## Sentiment
 
 Every color should convey a purpose, which is its sentiment
 
-### Neutral
+### <span class={css({ color: 'neutral.text.200' })}>Neutral</span>
 
-A palette for non-actionale items, eg. page content
+A palette for non-actionable items, eg. page content
 
-<a href='/reference/color/neutral-palette'>View the neutral color reference docs</a>
+<a class={textLink()} href="/reference/color/neutral-palette">
+  View the neutral color reference docs
+</a>
 
-### Action
+### <span class={css({ color: 'action.text.200' })}>Action</span>
 
 A palette that represents actionable and interactive items (eg. buttons, text links). Only used for status and feedback.
 
-<a href='/reference/color/action-palette'>View the action color reference docs</a>
+<a class={textLink()} href="/reference/color/action-palette">
+  View the action color reference docs
+</a>
 
-### Info
+### <span class={css({ color: 'info.text.200' })}>Info</span>
 
 A palette meant to communicate an informative message to the user (eg. Admonition, Toast). Only used for status and feedback.
 
-<a href='/reference/color/info-palette'>View the info color reference docs</a>
+<a class={textLink()} href="/reference/color/info-palette">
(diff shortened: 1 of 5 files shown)


docs/src/content/docs/explanation/color.mdx
```diff
@@ -5,44 +5,61 @@ hero:
...
 
+import { textLink } from '@/styled-system/recipes'
+import { css } from '@/styled-system/css'
+
 ## Sentiment
 
 Every color should convey a purpose, which is its sentiment
 
-### Neutral
+### <span class={css({ color: 'neutral.text.200' })}>Neutral</span>
```
caseybaggz, line 15: This feels like it would be a MD a11y error? If we are using JSX here I feel like we should just use JSX and not mix MD and JSX to have a cleaner markup for people who need it?

What do you think?
```suggestion
<h3 class={css({ color: 'neutral.text.200' })}>Neutral</h3>
```

bradryanbice (approved):

docs/src/content/docs/explanation/color.mdx
kathydlavoie-ps: Ironically this is what I had initially done - the reason I changed it to the way it is currently is that starlight seems to depends on the `###` to include in the right side nav

Screenshot 2024-01-18 at 1 42 43 PM

as for whether it is better to mix markdown & html, I am not entirely sure - the markdown `###` does actually translate to `h3` after compile, so I actually think it may just be ok. The html markup of this current structure comes out as 

Screenshot 2024-01-18 at 1 44 45 PM

which seems semantically appropriate. 

Additionally, using `h3` outright actually inherits starlight's colors which seem to override the pandacss function used - that may be something I can fix in the future, but it is a contributor to this structure. All in all, as far as a11y, I think the html structure comes out okay in the app - though I agree it does seem unusual. Is the a11y concern you have with the end experience or the dev experience?

caseybaggz (approved):
:shipit:

